PR #1, opened 2016-05-29
Added sf::Sprite and sf::Texture overload for ImageButton

susnux:
Maybe you are interested in my changes:
I have added sfml overloads for ImageButton (does not work otherwise for me).
Some other minor modifications, notable: If using ImGui::Image with only a sf::Sprite the the size of the sprite is taken (e.g. is uses now the scale of the sprite) I think for sprites this is more logical, for sf::Textures everything stays like it is.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -2,6 +2,7 @@
 #include <imgui.h>
 
 #include <SFML/OpenGL.hpp>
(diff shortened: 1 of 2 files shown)


EliasD:
Thank you! That's something I really needed myself, but wanted to implement later. 
Your overloads work as intended. And using sprite's global bounds is pretty neat idea.

PR #17, opened 2016-11-05
Fix #15: Set OpenGL Client state for color, texture, and coord arrays

### LINKED ISSUE #15: If you only draw ImGui stuff, it's not displayed if you don't call resetGLStates first
`resetGLStates`  is automatically called by SFML on the first `target.draw` call. It ensures that OpenGL state is set up correctly.

But I have no idea if calling `resetGLStates` on the first `RenderDrawLists` call is safe and won't mess everything up if you draw other stuff along with ImGui. If someone has ideas how to fix it correctly, please help. :D

iamPHEN:
Sets GL_VERTEX_ARRAY, GL_COLOR_ARRAY and GL_TEXTURE_COORD_ARRAY before rendering ImGUI elements.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -386,7 +386,9 @@ void RenderDrawLists(ImDrawData* draw_data)
             if (pcmd->UserCallback) {
                 pcmd->UserCallback(cmd_list, pcmd);
             } else {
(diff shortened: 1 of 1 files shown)


eliasdaler:
Somehow I already used all your changes and commited them to master branch. Maybe because I tested them individually and then forgot to merge instead. Sorry! But your contribution won't be forgotten. :D

PR #28, opened 2017-03-23
Update Image(texture, textureRect...)

rachelnertia:
Use abs on the size vector before passing it onwards for when given a texture rect with negative width and/or height.

That way, when you want to draw an image with 'flipped' texture coordinates, it won't end up to the left of where it ought to be, or above (ImGui::Image moves the 'cursor' by the size it is given, and draws from where it was to where it now is).


eliasdaler:
Great idea, merged. :)

PR #30, opened 2017-04-04
I added return !!!

Qiu3344:


eliasdaler:
Thanks, that's fair enough, but I added "return 0" as it's more commonly used in C++ and means the same thing most of the time. EXIT_SUCCESS also depends on <cstdlib> header.

Also, C++ standard says:
"If the value of status is zero or EXIT_SUCCESS, an implementation-defined form of the status successful termination is returned.". So, not much difference. Thanks again. :)

Qiu3344:
:)

2017-04-05 12:57 GMT+02:00 Elias Daler <notifications@github.com>:

PR #36, opened 2017-07-02
Update imgui-SFML.h

Epholys:
I'd like to merge this little modification.

When building my project I had some issues from multiple `#include` directives.
Adding `#pragma once` at the top of `imgui-SFML.h` solves this issue.

Thanks for this project!

### DIFF
--- a/imgui-SFML.h
+++ b/imgui-SFML.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <SFML/System/Vector2.hpp>
 #include <SFML/Graphics/Rect.hpp>
 #include <SFML/Graphics/Color.hpp>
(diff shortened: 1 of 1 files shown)


aggsol:
I would recommend using normal include guards for various reasons see: https://en.wikipedia.org/wiki/Pragma_once

Epholys:
Thanks for the tip @aggsol, you are right!

eliasdaler:
While I like `pragma once` more, I think that include guards are more universal, and I already had to abandon some things I like for portability. Thanks!

aggsol:
@eliasdaler as a benchmark I would say that every SFML does for prtability can't be that bad :-)

eliasdaler:
@aggsol agreed.

PR #37, opened 2017-08-31
Add android compatibility

Alia5:


eliasdaler:
Wow, that looks pretty neat, thanks!

I don't want to test it, I believe that it should work. Any screens of it running on Android? :)

Alia5:
Ahh, don't merge it yet. I just noticed that I shouldn't use ```#ifdef ANDROID``` since there are x86 Android devices with proper OpenGL...

I'll update quickly...

Will share screens when fixed ;)

eliasdaler:
Oops. :D
Should I revert it for now and merge it later?

Alia5:
Yup please...

PR #42, opened 2018-01-06
correct capitalization of GetFontTexture

dabbertorres:


eliasdaler:
Oops. That's a strange mistake to make. Thanks!

Issue #44, opened 2018-02-06
Adding "ImGui::CreateContext()" because of future breakage in imgui


Epholys:
There will be a huge merge to master in imgui : https://twitter.com/ocornut/status/960649767691980800. A little breaking change is that `ImGui::CreateContext()` will need to be called once. Does imgui-sfml need this modification?
![screenshot from 2018-02-06 22-28-12](https://user-images.githubusercontent.com/7680863/35885201-68378b7a-0b8d-11e8-98f7-d733114ae269.png)

eliasdaler:
Looks like ImGui::CreateContext() will be a thing in 1.60, and even though it's merged in ImGui's master, this version is still not released.
Once the 1.60 is released officially, I'll update ImGui-SFML. For now, I'll still support the latest release.

The only problem is that if someone just clones ImGui, ImGui-SFML won't work. What should I do? 🤔

Epholys:
I think the last solution is a good one, with just a little note in the readme to make it clear :)

eliasdaler:
Agreed! I'll do it soon then. :)

aggsol:
How about creating tags corresponding the tags of imgui like current master gets tag 1.53 and the next release gets 1.60 which it will be comptaible with?!

eliasdaler:
Oh no, entire month has passed! My definition of "soon" is close to Valve's...

Yes, this sounds good, but the problem with it is that I won't be able to patch bugs in 1.53 version if I find one when being on 1.60...

Alia5:
I guess we could do patch branches and merge them with squashed commits in that case.
But it wouldn't really be as clean.

SuperPrower:
I'm not sure if I'm in place in time, but I fixed all of my problems by replacing `ImGui::Shutdown` with `ImGui::DestroyContext` in imgui-SFML.cpp and running 'ImGui::CreateContext` in main.cpp before `ImGui::SFML::Init()`. Simple thing, but I think this may be useful for somebody.

eliasdaler:
Done. [Use this version if you use v.1.53!](https://github.com/eliasdaler/imgui-sfml/releases/tag/v.1.0)

nitro111:
anyád

PR #48, opened 2018-03-17
Fix issue #44

NoeMurr:
In the new version of ImGui the function ImGui::Shutdown() is private. That function is been replaced with ImGui::DestroyContext() and  ImGui::CreateContext() is been called in the ImGui::FSML::Init() function..


NoeMurr:
Done.

eliasdaler:
Done. Merged, thanks!

PR #50, opened 2018-03-31
fix too few arguments issue in DestroyContext

halcyonx:
According to v1.53 DestroyContext requires ImGuiContext


eliasdaler:
There's a note on the bottom of the README:

> Use ImGui-SFML's v1.0 if you're using ImGui's stable release (v.1.53)! This repo's master will be kept up to date with breaking changes in ImGui's master.

PR #52, opened 2018-04-10
Add README section about high DPI screens

JonnyPtn:
Includes a method to fix the issue on macOS. I'm not sure what solution to suggest for high DPI screens on other platforms.

I'll grab a screenshot of the issue for your blog post, but figured it's not worth adding to the README here?


eliasdaler:
Yeah, the text description is good enough. Thanks!

PR #59, opened 2018-07-07
Fix implicit cast warning

codecat:
Fixes this warning:

```
imgui.h(132): warning C4244: 'argument': conversion from 'const float' to 'int', possible loss of data
imgui-sfml.cpp(338): note: see reference to function template instantiation 'ImVec2::operator sf::Vector2<int>(void) const<int>' being compiled
```


codecat:
The `mousePos` vector is an `ImVec2`, does imgui-sfml provide an automatic cast for that?

eliasdaler:
Looks good, thanks! 

I should probably re-check all ImVec2 <-> sf::Vector2 conversions and not rely on implicit casts at all in ImGui-SFML's code...

PR #67, opened 2018-10-31
Add CACHE option to sets

Pinkii-:
Make it able to add subdirectory and use the variables from another cmake


eliasdaler:
Looks good, thanks!

PR #68, opened 2018-11-04
Add supression of warnings for Microsoft's compiler

GustavoQueipo:
Add supression of warnings (caused by converting from uint to void* in pCmd->TextureID) for Microsoft's compiler. Supression only worked for Clang and GCC.


eliasdaler:
Thanks a lot. :)

PR #73, opened 2019-01-27
Change the int/pointer casts to use memcpy.

relick:
- Add a static_assert to make sure we can safely put GLuint texture IDs into the ImTextureID (currently void*). Should catch cases where, _somehow_, GLuint is larger than the pointer size on the target system (and thus trying to use it for texture IDs would likely lead to memory corruption).

- Use memcpy to transfer data in and out of ImGui's texture IDs rather than cast. This method of type-punning is safe in regards to strict-aliasing (there was potential UB due to strict-aliasing rules at the original file's line 736, pointed out by `-Wstrict-aliasing`). It also allows us to remove the int-to-pointer cast warning suppression.

Code generated with optimisations should be identical or near identical as modern compilers recognise memcpy as a commonly used method of type-punning.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -555,7 +555,8 @@ void Image(const sf::Texture& texture,
 void Image(const sf::Texture& texture, const sf::Vector2f& size,
     const sf::Color& tintColor, const sf::Color& borderColor)
 {
-    ImGui::Image(convertGLTextureHandleToImTextureID(texture.getNativeHandle()), size, ImVec2(0, 0), ImVec2(1, 1), tintColor, borderColor);
+    ImTextureID textureID = convertGLTextureHandleToImTextureID(texture.getNativeHandle());
+    ImGui::Image(textureID, size, ImVec2(0, 0), ImVec2(1, 1), tintColor, borderColor);
 }
 
 void Image(const sf::Texture& texture, const sf::FloatRect& textureRect,
@@ -571,7 +572,9 @@ void Image(const sf::Texture& texture, const sf::Vector2f& size, const sf::Float
     ImVec2 uv0(textureRect.left / textureSize.x, textureRect.top / textureSize.y);
     ImVec2 uv1((textureRect.left + textureRect.width) / textureSize.x,
         (textureRect.top + textureRect.height) / textureSize.y);
-    ImGui::Image(convertGLTextureHandleToImTextureID(texture.getNativeHandle()), size, uv0, uv1, tintColor, borderColor);
+
+    ImTextureID textureID = convertGLTextureHandleToImTextureID(texture.getNativeHandle());
+    ImGui::Image(textureID, size, uv0, uv1, tintColor, borderColor);
 }
 
 void Image(const sf::Sprite& sprite,
@@ -748,7 +751,8 @@ void RenderDrawLists(ImDrawData* draw_data)
             if (pcmd->UserCallback) {
                 pcmd->UserCallback(cmd_list, pcmd);
             } else {
-                glBindTexture(GL_TEXTURE_2D, convertImTextureIDToGLTextureHandle(pcmd->TextureId));
+                GLuint textureHandle = convertImTextureIDToGLTextureHandle(pcmd->TextureId);
+                glBindTexture(GL_TEXTURE_2D, textureHandle);
                 glScissor((int)pcmd->ClipRect.x, (int)(fb_height - pcmd->ClipRect.w),
                     (int)(pcmd->ClipRect.z - pcmd->ClipRect.x), (int)(pcmd->ClipRect.w - pcmd->ClipRect.y));
                 glDrawElements(GL_TRIANGLES, (GLsizei)pcmd->ElemCount, GL_UNSIGNED_SHORT, idx_buffer);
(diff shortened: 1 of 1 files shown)


imgui-SFML.cpp
```diff
@@ -114,15 +115,7 @@ int closeKeyboardIME()
...
-// Supress warnings caused by converting from uint to void* in pCmd->TextureID
-#ifdef _MSC_VER
-#pragma warning(push)
-#pragma warning(disable:4312)
-#elif __clang__
-#pragma clang diagnostic ignored "-Wint-to-void-pointer-cast" // warning : cast to 'void *' from smaller integer type 'int'
-#elif defined(__GNUC__)
-#pragma GCC diagnostic ignored "-Wint-to-pointer-cast"      // warning: cast to pointer from integer of different size
-#endif
+static_assert(sizeof(GLuint) <= sizeof(ImTextureID), "ImTextureID is not large enough to fit GLuint.");
```
eliasdaler: static_assert is a C++11 feature... This repo is using C++03. I think it'll be okay to put this assert under a macro which checks if we're in C++>=11 and to a runtime assert in ImGui::SFML::Init

imgui-SFML.cpp
```diff
@@ -553,7 +547,11 @@ void Image(const sf::Texture& texture,
 void Image(const sf::Texture& texture, const sf::Vector2f& size,
     const sf::Color& tintColor, const sf::Color& borderColor)
 {
-    ImGui::Image((void*)texture.getNativeHandle(), size, ImVec2(0, 0), ImVec2(1, 1), tintColor, borderColor);
+    GLuint nativeHandle = texture.getNativeHandle();
+    ImTextureID textureID;
+    std::memcpy(&textureID, &nativeHandle, sizeof(GLuint));
```
eliasdaler: These three lines are repeated in quite a few places now... Let's make some function: `ImTextureID convertToImGuiTextureHandle(GLuint)`

Plus, don't forget to initialize ImTextureID there (to `NULL`, I guess...), otherwise some bits can be garbage when `sizeof(GLUint) < sizeof(ImTextureId)`

imgui-SFML.cpp
```diff
@@ -733,7 +736,8 @@ void RenderDrawLists(ImDrawData* draw_data)
             if (pcmd->UserCallback) {
                 pcmd->UserCallback(cmd_list, pcmd);
             } else {
-                GLuint tex_id = (GLuint)*((unsigned int*)&pcmd->TextureId);
+                GLuint tex_id;
+                std::memcpy(&tex_id, &pcmd->TextureId, sizeof(GLuint));
```
eliasdaler: And here we'll need `GLuint convertToGLTextureHandle(ImTextureID)` to make this easier for future readings of code :)

imgui-SFML.cpp
```diff
@@ -114,15 +115,7 @@ int closeKeyboardIME()
...
 
-// Supress warnings caused by converting from uint to void* in pCmd->TextureID
-#ifdef _MSC_VER
-#pragma warning(push)
-#pragma warning(disable:4312)
-#elif __clang__
-#pragma clang diagnostic ignored "-Wint-to-void-pointer-cast" // warning : cast to 'void *' from smaller integer type 'int'
-#elif defined(__GNUC__)
-#pragma GCC diagnostic ignored "-Wint-to-pointer-cast"      // warning: cast to pointer from integer of different size
-#endif
```
eliasdaler: Really awesome that we can remove it now!

imgui-SFML.cpp
relick: I've added the helper function - I don't believe the initialisation of `ImTextureID` is necessary as we never read from it directly, only ever memcpy (which is always valid) back into a `GLuint` which we then read from. Thus, even with garbage, I don't think this is UB (nor would it lead to inconsistent read/writes as we always use `sizeof(GLuint)`). I am hesitant to initialise it as when I checked on godbolt, initialising seemed to add an extra instruction on the latest gcc.

relick: On second thoughts, the copying of the ImTextureIDs around likely counts as a direct read so I'll go ahead and add the initialisation.

imgui-SFML.cpp
```diff
@@ -547,11 +555,7 @@ void Image(const sf::Texture& texture,
 void Image(const sf::Texture& texture, const sf::Vector2f& size,
     const sf::Color& tintColor, const sf::Color& borderColor)
 {
-    GLuint nativeHandle = texture.getNativeHandle();
-    ImTextureID textureID;
-    std::memcpy(&textureID, &nativeHandle, sizeof(GLuint));
-
-    ImGui::Image(textureID, size, ImVec2(0, 0), ImVec2(1, 1), tintColor, borderColor);
+    ImGui::Image(convertGLTextureHandleToImTextureID(texture.getNativeHandle()), size, ImVec2(0, 0), ImVec2(1, 1), tintColor, borderColor);
```
eliasdaler: Can you make separate variables for textureID here and in other calls? Right now it's a bit hard to read
After that, I'll merge :)

relick: No problem!

eliasdaler:
Looks good, thanks a lot. :)

Issue #81, opened 2019-05-15
Make ImGui a subrepo of ImGui-SFML


eliasdaler:
I think it'll be good to pin specific ImGui commit (I'll use release ones in "master" and releases) and allow to not specify IMGUI_DIR, but get ImGui from subrepository and be sure that the version you'll be using was tested with ImGui-SFML.

RichardDally:
It brings more complexity for almost no gain, I prefer cloning imgui separately on my own.

jdoubleu:
A version table would be nice though. E.g. ImGui-SFML v2.5.1 is compatible with SFML v2.5.1 and imGui v1.xx

Alia5:
I guess you mean submodule?   
Careful with the wording here 😅

In any case, I'd strongly argue for a submodule instead of git-subtree or [subrepo](https://github.com/ingydotnet/git-subrepo) **and** make sure that the path of ImGUI is configurable.
So don't take `IMGUI_DIR` away, but provide a default for a (may or may not) cloned submodule.

This way, folks who want to manually manage ImGUI can do so; For other folks, the submodule should be enough and works out of the box.

rezzeted:
You can't add the sources of another library to the sources of one library as a git submodule. This way you will link two libraries at the level of their source code. At first glance, this seems like a good idea, but think about how it will work in various package managers that build libraries from source code, such as vcpkg? After all, the package repository already contains imgui as a separate library, and which version of the source code should be used in this case will conflict.
The compatibility of versions should be thought about by the people maintaining the package manager repository, although this often leads to problems.

PR #82, opened 2019-05-18
Conanfile for packaging

barsoosayque:
Introducing `conanfile.py` for ImGui-SFML to ease usage for people who want to use it as a dependency in a conan project. 

Using `conan create . user/stable` is enough to build a package localy, however, it would be much nicer to [submit an inclusion request](https://docs.conan.io/en/latest/uploading_packages/bintray/conan_center_guide.html) to **conan-center**. This way, people can just use *imgui-sfml/2.0.1@conan/stable* or something like that in their recipies. Another alternative is **Bincrafters**, although I'm not familiar with their inclusion politics.

I didn't change README.md to reflect new conan package because at the moment I don't know at what conan remote, channel or user this package will be. Also, because of that, I'm marking this PR as a draft to discuss the situation.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -56,7 +56,7 @@ class ImguiSfmlConan(ConanFile):
         'imgui_revision': 'v1.70'
     }
     exports_sources = [
-        'cmake/FindImGui.cmake'
+        'cmake/FindImGui.cmake',
         'CMakeLists.txt',
         'imconfig-SFML.h',
         'imgui-SFML.cpp',
--- a/conanfile.py
+++ b/conanfile.py
@@ -26,9 +26,9 @@ class ImguiSFML(ConanFile):
       Use 'None' if you want this build to include imconfig.h file,
       otherwise use anything else.
 
-    * imgui_revision: None or String
+    * imgui_revision: String
       Tag or branch of ImGui repository that should be used with
-      ImGui-SFML. 'None' will results in master branch.
+      ImGui-SFML.
     """
 
     name = 'ImGui-SFML'
@@ -67,14 +67,16 @@ class ImguiSFML(ConanFile):
     _imgui_dir = 'imgui'
 
     def source(self):
-        imgui_revision = str(self.options.imgui_revision) if self.options.imgui_revision else 'master'
         git = Git(folder=self._imgui_dir)
-        git.clone('https://github.com/ocornut/imgui.git', imgui_revision)
+        git.clone('https://github.com/ocornut/imgui.git', str(self.options.imgui_revision))
 
     def configure(self):
         imconfig = self.options.imconfig
         if imconfig and not os.path.isfile(str(imconfig)):
             raise ConanInvalidConfiguration("Provided ImGui config is not a file or doesn't exist")
+        if not self.options.imgui_revision:
+            raise ConanInvalidConfiguration("ImGui revision is empty. Try latest version tag or 'master'")
+
         self.options['sfml'].graphics = True
         self.options['sfml'].window = True
         self.options['sfml'].shared = self.options.shared
--- a/conanfile.py
+++ b/conanfile.py
@@ -38,7 +38,7 @@ class ImguiSFML(ConanFile):
     url = 'https://github.com/eliasdaler/imgui-sfml'
     homepage = 'https://github.com/eliasdaler/imgui-sfml'
     author = 'Elias Daler <eliasdaler@yandex.ru>'
-    build_requires = 'sfml/2.5.1@bincrafters/stable'
+    requires = 'sfml/2.5.1@bincrafters/stable'
     license = 'MIT'
     settings = 'os', 'compiler', 'build_type', 'arch'
     options = {
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import os.path
 from conans import ConanFile, CMake
 from conans.tools import Git
-import os
 
 class ImguiSFML(ConanFile):
     """ Imgui-SFML library conan file.
@@ -29,10 +29,10 @@ class ImguiSFML(ConanFile):
     name = 'ImGui-SFML'
     version = '2.0.1'
     description = 'ImGui binding for use with SFML'
-    topics = ('conan', 'sfml', 'imgui')
+    topics = ('conan', 'sfml', 'gui', 'imgui')
     url = 'https://github.com/eliasdaler/imgui-sfml'
     homepage = 'https://github.com/eliasdaler/imgui-sfml'
-    author = 'eliasdaler <eliasdaler@yandex.ru>'
+    author = 'Elias Daler <eliasdaler@yandex.ru>'
     build_requires = 'sfml/2.5.1@bincrafters/stable'
     license = 'MIT'
     settings = 'os', 'compiler', 'build_type', 'arch'
@@ -48,7 +48,8 @@ class ImguiSFML(ConanFile):
         'imconfig': None,
         'imconfig_install_folder': None
     }
-    exports_sources = ['*.cpp', '*.h', 'CMakeLists.txt', 'cmake*']
+    exports_sources = ['imgui-SFML.cpp', 'imgui-SFML.h', 'imconfig-SFML.h',
+                       'imgui-SFML_export.h', 'CMakeLists.txt', 'cmake/FindImGui.cmake']
     exports = 'LICENSE.md'
     _imgui_dir = 'imgui'
 
@@ -59,9 +60,9 @@ class ImguiSFML(ConanFile):
     def configure(self):
         self.options['sfml'].graphics = True
         self.options['sfml'].window = True
-        self.options['sfml'].shared = True
+        self.options['sfml'].shared = self.options.shared
 
-    def configure_cmake(self):
+    def _configure_cmake(self):
         cmake = CMake(self)
         cmake.definitions['IMGUI_DIR'] = os.path.join(self.source_folder, self._imgui_dir)
         cmake.definitions['SFML_DIR'] = os.path.join(self.deps_cpp_info['sfml'].lib_paths[0], 'cmake', 'SFML')
@@ -80,11 +81,11 @@ class ImguiSFML(ConanFile):
         return cmake
 
     def build(self):
-        cmake = self.configure_cmake()
+        cmake = self._configure_cmake()
         cmake.build()
 
     def package(self):
-        cmake = self.configure_cmake()
+        cmake = self._configure_cmake()
         cmake.install()
 
     def package_info(self):
--- a/conanfile.py
+++ b/conanfile.py
@@ -12,10 +12,10 @@ class ImguiSfmlConan(ConanFile):
     Options:
 
     * shared: True or False
-      Whether to build this library shared or static
+      Whether to build this library shared or static.
 
     * fPIC: True or False
-      Whether or not to use -fPIC option while building
+      Whether or not to use -fPIC option while building.
 
     * imconfig: None or String
       Use 'None' if you want this library to provide default
@@ -63,7 +63,7 @@ class ImguiSfmlConan(ConanFile):
         'imgui-SFML_export.h',
         'imgui-SFML.h',
     ]
-    exports = 'LICENSE.md'
+    exports = 'LICENSE'
     _imgui_dir = 'imgui'
 
     def source(self):
@@ -74,6 +74,8 @@ class ImguiSfmlConan(ConanFile):
         imconfig = self.options.imconfig
         if imconfig and not os.path.isfile(str(imconfig)):
             raise ConanInvalidConfiguration("Provided ImGui config is not a file or doesn't exist")
+        else:
+            self.exports_sources.append(str(imconfig))
         if not self.options.imgui_revision:
             raise ConanInvalidConfiguration("ImGui revision is empty. Try latest version tag or 'master'")
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -4,6 +4,7 @@
 import os.path
 from conans import ConanFile, CMake
 from conans.tools import Git
+from conans.errors import ConanInvalidConfiguration
 
 class ImguiSFML(ConanFile):
     """ Imgui-SFML library conan file.
@@ -58,6 +59,9 @@ class ImguiSFML(ConanFile):
         git.clone('https://github.com/ocornut/imgui.git', 'v1.69')
 
     def configure(self):
+        imconfig = self.options.imconfig
+        if imconfig and not os.path.isfile(str(imconfig)):
+            raise ConanInvalidConfiguration("Provided ImGui config is not a file or does not exists")
         self.options['sfml'].graphics = True
         self.options['sfml'].window = True
         self.options['sfml'].shared = self.options.shared
--- a/conanfile.py
+++ b/conanfile.py
@@ -23,8 +23,9 @@ class ImguiSfmlConan(ConanFile):
       imconfig.h file you want to include in the build.
 
     * imconfig_install_folder: None or String
-      Use 'None' if you want this build to include imconfig.h file,
-      otherwise use anything else.
+      Use 'None' if you want this build to include custom config
+      file, otherwise specify the directory where your imconfig
+      will be installed.
 
     * imgui_revision: String
       Tag or branch of ImGui repository that should be used with
@@ -72,10 +73,11 @@ class ImguiSfmlConan(ConanFile):
 
     def configure(self):
         imconfig = self.options.imconfig
-        if imconfig and not os.path.isfile(str(imconfig)):
-            raise ConanInvalidConfiguration("Provided ImGui config is not a file or doesn't exist")
-        else:
-            self.exports_sources.append(str(imconfig))
+        if imconfig:
+            if not os.path.isfile(str(imconfig)):
+                raise ConanInvalidConfiguration("Provided user config is not a file or doesn't exist")
+            else:
+                self._imconfig_path = os.path.abspath(str(self.options.imconfig))
         if not self.options.imgui_revision:
             raise ConanInvalidConfiguration("ImGui revision is empty. Try latest version tag or 'master'")
 
@@ -92,10 +94,9 @@ class ImguiSfmlConan(ConanFile):
         if self.options.imconfig_install_folder:
             cmake.definitions['IMGUI_SFML_CONFIG_INSTALL_DIR'] = self.options.imconfig_install_folder
         if self.options.imconfig:
-            imconfig = str(self.options.imconfig)
             cmake.definitions['IMGUI_SFML_USE_DEFAULT_CONFIG'] = 'OFF'
-            cmake.definitions['IMGUI_SFML_CONFIG_NAME'] = os.path.basename(imconfig)
-            cmake.definitions['IMGUI_SFML_CONFIG_DIR'] = os.path.dirname(imconfig)
+            cmake.definitions['IMGUI_SFML_CONFIG_NAME'] = os.path.basename(self._imconfig_path)
+            cmake.definitions['IMGUI_SFML_CONFIG_DIR'] = os.path.dirname(self._imconfig_path)
         else:
             cmake.definitions['IMGUI_SFML_USE_DEFAULT_CONFIG'] = 'ON'
         cmake.configure()
--- a/conanfile.py
+++ b/conanfile.py
@@ -33,7 +33,7 @@ class ImguiSfmlConan(ConanFile):
     """
 
     name = 'imgui-sfml'
-    version = '2.0.1'
+    version = '2.0.2'
     description = 'ImGui binding for use with SFML'
     topics = ('conan', 'sfml', 'gui', 'imgui')
     url = 'https://github.com/eliasdaler/imgui-sfml'
--- a/conanfile.py
+++ b/conanfile.py
@@ -25,6 +25,10 @@ class ImguiSFML(ConanFile):
     * imconfig_install_folder: None or String
       Use 'None' if you want this build to include imconfig.h file,
       otherwise use anything else.
+
+    * imgui_revision: None or String
+      Tag or branch of ImGui repository that should be patched with
+      ImGui-SFML. 'None' will results in master branch.
     """
 
     name = 'ImGui-SFML'
@@ -41,13 +45,15 @@ class ImguiSFML(ConanFile):
         'shared': [True, False],
         'fPIC': [True, False],
         'imconfig': 'ANY',
-        'imconfig_install_folder': 'ANY'
+        'imconfig_install_folder': 'ANY',
+        'imgui_revision': 'ANY'
     }
     default_options = {
         'shared': False,
         'fPIC': True,
         'imconfig': None,
-        'imconfig_install_folder': None
+        'imconfig_install_folder': None,
+        'imgui_revision': 'v1.70'
     }
     exports_sources = [
         'imgui-SFML.cpp',
@@ -61,8 +67,9 @@ class ImguiSFML(ConanFile):
     _imgui_dir = 'imgui'
 
     def source(self):
+        imgui_revision = str(self.options.imgui_revision) if self.options.imgui_revision else 'master'
         git = Git(folder=self._imgui_dir)
-        git.clone('https://github.com/ocornut/imgui.git', 'v1.69')
+        git.clone('https://github.com/ocornut/imgui.git', imgui_revision)
 
     def configure(self):
         imconfig = self.options.imconfig
--- a/conanfile.py
+++ b/conanfile.py
@@ -49,8 +49,14 @@ class ImguiSFML(ConanFile):
         'imconfig': None,
         'imconfig_install_folder': None
     }
-    exports_sources = ['imgui-SFML.cpp', 'imgui-SFML.h', 'imconfig-SFML.h',
-                       'imgui-SFML_export.h', 'CMakeLists.txt', 'cmake/FindImGui.cmake']
+    exports_sources = [
+        'imgui-SFML.cpp',
+        'imgui-SFML.h',
+        'imconfig-SFML.h',
+        'imgui-SFML_export.h',
+        'CMakeLists.txt',
+        'cmake/FindImGui.cmake'
+    ]
     exports = 'LICENSE.md'
     _imgui_dir = 'imgui'
 
@@ -61,7 +67,7 @@ class ImguiSFML(ConanFile):
     def configure(self):
         imconfig = self.options.imconfig
         if imconfig and not os.path.isfile(str(imconfig)):
-            raise ConanInvalidConfiguration("Provided ImGui config is not a file or does not exists")
+            raise ConanInvalidConfiguration("Provided ImGui config is not a file or doesn't exist")
         self.options['sfml'].graphics = True
         self.options['sfml'].window = True
         self.options['sfml'].shared = self.options.shared
--- a/conanfile.py
+++ b/conanfile.py
@@ -6,7 +6,7 @@ from conans import ConanFile, CMake
 from conans.tools import Git
 from conans.errors import ConanInvalidConfiguration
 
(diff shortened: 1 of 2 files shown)


conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+
+    * imconfig_install_folder: None or String
+      Use 'None' if you want this build to include imconfig.h file,
+      otherwise use anything else.
+    """
+
+    name = 'ImGui-SFML'
+    version = '2.0.1'
+    description = 'ImGui binding for use with SFML'
+    topics = ('conan', 'sfml', 'imgui')
```
eliasdaler: Is "Conan" a really good topic here?

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+    url = 'https://github.com/eliasdaler/imgui-sfml'
+    homepage = 'https://github.com/eliasdaler/imgui-sfml'
+    author = 'eliasdaler <eliasdaler@yandex.ru>'
```
eliasdaler: Should be "Elias Daler", not "eliasdaler"

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+        'imconfig': None,
+        'imconfig_install_folder': None
+    }
+    exports_sources = ['*.cpp', '*.h', 'CMakeLists.txt', 'cmake*']
+    exports = 'LICENSE.md'
+    _imgui_dir = 'imgui'
+
+    def source(self):
+        git = Git(folder=self._imgui_dir)
+        git.clone('https://github.com/ocornut/imgui.git', 'v1.69')
```
eliasdaler: Not very nice... This should be a parameter too, I think.

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def configure(self):
+        self.options['sfml'].graphics = True
+        self.options['sfml'].window = True
+        self.options['sfml'].shared = True
```
eliasdaler: This should depend on whether ImGui-SFML is build as static or shared lib

conanfile.py
eliasdaler: Also, need to check that clone succeeded and report error if not

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+        'imconfig': 'ANY',
+        'imconfig_install_folder': 'ANY'
+    }
+    default_options = {
+        'shared': False,
+        'fPIC': True,
+        'imconfig': None,
+        'imconfig_install_folder': None
+    }
+    exports_sources = ['*.cpp', '*.h', 'CMakeLists.txt', 'cmake*']
```
eliasdaler: It's better to explicitly list files there and not use globs. There's not many of them.

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+    def configure(self):
+        self.options['sfml'].graphics = True
+        self.options['sfml'].window = True
+        self.options['sfml'].shared = True
+
+    def configure_cmake(self):
+        cmake = CMake(self)
+        cmake.definitions['IMGUI_DIR'] = os.path.join(self.source_folder, self._imgui_dir)
+        cmake.definitions['SFML_DIR'] = os.path.join(self.deps_cpp_info['sfml'].lib_paths[0], 'cmake', 'SFML')
+        cmake.definitions['IMGUI_SFML_BUILD_EXAMPLES'] = 'OFF'
```
eliasdaler: Should be configurable?

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+        cmake.definitions['IMGUI_SFML_FIND_SFML'] = 'ON'
```
eliasdaler: Should be "OFF" if you're specifying "SFML_DIR"

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+        if self.options.imconfig_install_folder:
+            cmake.definitions['IMGUI_SFML_CONFIG_INSTALL_DIR'] = self.options.imconfig_install_folder
+        if self.options.imconfig:
+            imconfig = str(self.options.imconfig)
+            cmake.definitions['IMGUI_SFML_USE_DEFAULT_CONFIG'] = 'OFF'
```
eliasdaler: Would be nice to check that user-specified imconfig exists here

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+            cmake.definitions['IMGUI_SFML_CONFIG_NAME'] = os.path.basename(imconfig)
+            cmake.definitions['IMGUI_SFML_CONFIG_DIR'] = os.path.dirname(imconfig)
+        else:
+            cmake.definitions['IMGUI_SFML_USE_DEFAULT_CONFIG'] = 'ON'
```
eliasdaler: This is "ON" by default, not really needed

conanfile.py
eliasdaler: Btw, don't really like that we need to do "git.clone" here... I think I'll go with subrepo approach soon. See #81

conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+    def source(self):
+        git = Git(folder=self._imgui_dir)
+        git.clone('https://github.com/ocornut/imgui.git', 'v1.69')
+
+    def configure(self):
+        self.options['sfml'].graphics = True
+        self.options['sfml'].window = True
+        self.options['sfml'].shared = True
+
+    def configure_cmake(self):
```
eliasdaler: Not a public interface, name it `_configure_cmake`

conanfile.py
eliasdaler: I'd also add "GUI" and maybe "game development" topic

conanfile.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from conans import ConanFile, CMake
+from conans.tools import Git
+import os
```
eliasdaler: You don't need whole "os" imported here. Just import "os.path"

eliasdaler:
> If you are the author of an open source library, your package will be approved. Keep in mind that it is your responsibility to maintain acceptable standards of quality for all packages you submit for inclusion in conan-center.

This makes the situation pretty easy. Let's finish this PR and merge it into ImGui-SFML (maybe even name it 2.0.2). Then, I'll submit a request to conan-center.

conanfile.py
barsoosayque: I took [**conan-sfml**](https://github.com/bincrafters/conan-sfml/blob/stable/2.5.1/conanfile.py) as a main example and I also checked **conan-openal** (both from bincrafters), and they do have 'conan' in topics. Though, there is no 'conan' tag at respective packages in bintray.

+1 for GUI, though.

conanfile.py
barsoosayque: Ah, sorry. Will fix.

conanfile.py
barsoosayque: > Also, need to check that clone succeeded and report error if not

Regarding the success of cloning: it emits errors automatically and stops the whole build.

> This should be a parameter too, I think.

Can you please elaborate on this ? I just see ImGui as a dependency, so hardcoding its version is a safe way to establish future builds.

> I think I'll go with subrepo approach soon.

Aha, I saw that, and since it's `v3.0`, I just decided to do it the old way. :slightly_smiling_face:
Second thought, is subrepo technically any different from current `git.clone` ?

conanfile.py
barsoosayque: Since conan packages are primary used as dependencies, there is no need for that. I looked up a similar case in **conan-openal**, and they just hardcoded examples as `False`:

https://github.com/bincrafters/conan-openal/blob/a01c25eed47685ff25bd6b98297fd189e7aabb82/conanfile.py#L44-L46

conanfile.py
barsoosayque: I had errors because of doing so. Let me check if this is working now..

conanfile.py
barsoosayque: Yup. However, I believe having that line is quite a verbose way to state that cmake *will* use the default config (if someone inspecting conan recipe from the command line, there is such a feature). It's more about understanding recipe at glance.

conanfile.py
barsoosayque: Yes. If I set "SFML_DIR" and disable "IMGUI_SFML_FIND_SFML", then I get compile errors:

```
../imgui-SFML.h:4:10: fatal error: 'SFML/System/Vector2.hpp' file not found
#include <SFML/System/Vector2.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~
```

And some hints from CMake:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_EXPORT_NO_PACKAGE_REGISTRY
    CONAN_CMAKE_POSITION_INDEPENDENT_CODE
    CONAN_COMPILER
    CONAN_COMPILER_VERSION
    CONAN_CXX_FLAGS
    CONAN_C_FLAGS
    CONAN_EXPORTED
    CONAN_IN_LOCAL_CACHE
    CONAN_LIBCXX
    CONAN_SHARED_LINKER_FLAGS
    SFML_DIR
```

conanfile.py
eliasdaler: >  I just see ImGui as a dependency, so hardcoding its version is a safe way to establish future builds

Then let's go with the latest possible ImGui version that ImGui-SFML can handle (1.70).
And the problem with hardcoding it is that ImGui releases more frequently than ImGui-SFML, so people might want to use newer ImGui with ImGui-SFML, but unable to do so.

As for subrepo - yes, but it'll allow me to "pin" ImGui version and clone exactly needed one unless specified otherwise. This will allow Conan script to be updated less often. Let's leave "git.clone" as is for now, but add ImGui version parameter.

conanfile.py
eliasdaler: I see. Are you sure that CMake finds the version (via `find_package` in ImGui-SFML's CMakeLists.txt) that Conan downloads because of dependency? If so, then there's no problem, though I don't see why you need to set 'SFML_DIR' in that case.

conanfile.py
eliasdaler: Okay, fair enough.

conanfile.py
```diff
@@ -48,7 +49,8 @@ class ImguiSFML(ConanFile):
         'imconfig': None,
         'imconfig_install_folder': None
     }
-    exports_sources = ['*.cpp', '*.h', 'CMakeLists.txt', 'cmake*']
+    exports_sources = ['imgui-SFML.cpp', 'imgui-SFML.h', 'imconfig-SFML.h',
```
eliasdaler: Minor, but let's not do it in one line.
One line per file, please. Will be easier for git later.

conanfile.py
```diff
@@ -57,11 +59,14 @@ def source(self):
         git.clone('https://github.com/ocornut/imgui.git', 'v1.69')
 
     def configure(self):
+        imconfig = self.options.imconfig
+        if imconfig and not os.path.isfile(str(imconfig)):
+            raise ConanInvalidConfiguration("Provided ImGui config is not a file or does not exists")
```
eliasdaler: "doesn't exist"*

conanfile.py
barsoosayque: > Are you sure that CMake finds the version <...> that Conan downloads because of dependency?

Yeah. CMake logs confirm that:

```
-- Found SFML 2.5.1 in ~/.conan/data/sfml/2.5.1/bincrafters/stable/package/ed8c36dd1caf5fec34121c6dc29badcbf3252fab/lib/cmake/SFML
-- Found ImGui v1.69 in ~/.conan/data/ImGui-SFML/2.0.1/user/stable/build/eb6bb74157d96eac87398f1d4b33528edcded47f/imgui
```

---

> I don't see why you need to set 'SFML_DIR' in that case.

Hrm. My bet is that's what `find_package` wants from me. With `IMGUI_SFML_FIND_SFML=ON` and no `SFML_DIR` I got this:

``` 
Could not find a package configuration file provided by "SFML" (requested
version 2.5) with any of the following names:

  SFMLConfig.cmake
  sfml-config.cmake

Add the installation prefix of "SFML" to CMAKE_PREFIX_PATH or set
"SFML_DIR" to a directory containing one of the above files.  If "SFML"
provides a separate development package or SDK, be sure it has been
installed.
```

And the content of aformentioned `SFML_DIR`:

```bash
$ ls ~/.conan/data/sfml/2.5.1/bincrafters/stable/package/ed8c36dd1caf5fec34121c6dc29badcbf3252fab/lib/cmake/SFML
SFMLConfig.cmake  SFMLConfigDependencies.cmake  SFMLConfigVersion.cmake  SFMLSharedTargets-release.cmake  SFMLSharedTargets.cmake
```

eliasdaler: Oh, now I understand how this works. You're getting pre-build SFML, of course, so it contains SFMLConfig.cmake and you need to point CMake to it. Okay, looks good

conanfile.py
```diff
@@ -25,6 +25,10 @@ class ImguiSFML(ConanFile):
     * imconfig_install_folder: None or String
       Use 'None' if you want this build to include imconfig.h file,
       otherwise use anything else.
+
+    * imgui_revision: None or String
+      Tag or branch of ImGui repository that should be patched with
```
eliasdaler: "should be pached" => "should be used"?

conanfile.py
```diff
@@ -25,6 +25,10 @@ class ImguiSFML(ConanFile):
...
+      ImGui-SFML. 'None' will results in master branch.
```
eliasdaler: Let's drop this behaviour. If the user specifies 'None', they can easily specify 'master'

conanfile.py
```diff
@@ -0,0 +1,109 @@
...
+        'imconfig_install_folder': None,
+        'imgui_revision': 'v1.70'
+    }
+    exports_sources = [
+        'imgui-SFML.cpp',
+        'imgui-SFML.h',
+        'imconfig-SFML.h',
+        'imgui-SFML_export.h',
+        'CMakeLists.txt',
+        'cmake/FindImGui.cmake'
```
eliasdaler: I'd like to have this order (lexicographical sort with items in subdirs first):

```python
'cmake/FindImGui.cmake',
'CMakeLists.txt',  
'imconfig-SFML.h',
'imgui-SFML.cpp',
'imgui-SFML_export.h',
'imgui-SFML.h'
 ```

File: conanfile.py [hunk 17]
```diff
@@ -0,0 +1,109 @@
...
+    ]
+    exports = 'LICENSE.md'
+    _imgui_dir = 'imgui'
+
+    def source(self):
+        imgui_revision = str(self.options.imgui_revision) if self.options.imgui_revision else 'master'
```
eliasdaler: Let's make it an error to set `imgui_revision` to `None`.
It'll force users to set it to 'master' if they want, which will be more predictive.

conanfile.py
```diff
@@ -0,0 +1,109 @@
...
+    def build(self):
+        cmake = self._configure_cmake()
+        cmake.build()
+
+    def package(self):
+        cmake = self._configure_cmake()
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ['ImGui-SFML']
```
eliasdaler: Why is extension omitted here?
Plus, on Linux this will be 'libImGui-SFML.a' or `libImGui-SFML.so`

eliasdaler: Also, if ImGui-SFML was build as shared lib, user should also link to SFML libs, so they should be also added to ```self.cpp_info.libs```, I think.

barsoosayque: > Why is extension omitted here?

It is automatically managed.

> <...> user should also link to SFML libs

They will. I just gotta change sfml from `build_requires` to regular `requires`.

conanfile.py
barsoosayque: As per current implementation, ImGui-SFML providing ImGui with SFML patches. I wanted to make it clear for users, but I don't have a strong opinion on that, because it's basically just wording. I will change that, of course, if you find the second one more suitable.

eliasdaler: There are no "patches" in ImGui's source code. imconfig-SFML.h is an additional header which is included via the standard extension mechanism.

conanfile.py
```diff
@@ -0,0 +1,111 @@
...
+
+class ImguiSFML(ConanFile):
+    """ Imgui-SFML library conan file.
+
+    Options:
+
+    * shared: True or False
+      Whether to build this library shared or static
+
+    * fPIC: True or False
```
eliasdaler: Is this still needed?

barsoosayque: `fPIC` option ? I've seen it in every library recipe I used so far, and just like `shared` option, it is automatically managed with CMake.

eliasdaler: I see. Seems like Conan passes it automatically somehow..,

eliasdaler:
On what platforms and configurations did you test this recipe? Also, which Conan version should I use for testing this recipe?

barsoosayque:
> On what platforms and configurations did you test this recipe?

ArchLinux with clang and Windows 10 with msvs. I didn't test the latest changes on Windows, though, only the initial commit.

> Also, which Conan version should I use for testing this recipe?

Ah.. It's hard to say. I've been using 1.15.0, and if I recall correctly, I had to update my conan to use some feature there.

eliasdaler:
By the way, will the package be called 'ImGui-SFML'? Maybe we should go with 'imgui-sfml', seeing how 'SFML' package is named 'sfml'? Looks like it's common for other packages to be named like in lower case..

barsoosayque:
Yeah, seems to be reasonable with the consistency of the packages. I missed that, thanks. Gotta change that..

conanfile.py
```diff
@@ -0,0 +1,111 @@
...
+    }
+    default_options = {
+        'shared': False,
+        'fPIC': True,
+        'imconfig': None,
+        'imconfig_install_folder': None,
+        'imgui_revision': 'v1.70'
+    }
+    exports_sources = [
+        'cmake/FindImGui.cmake'
```
eliasdaler: Missing comma

conanfile.py
```diff
@@ -0,0 +1,111 @@
...
+            raise ConanInvalidConfiguration("ImGui revision is empty. Try latest version tag or 'master'")
+
+        self.options['sfml'].graphics = True
+        self.options['sfml'].window = True
+        self.options['sfml'].shared = self.options.shared
+
+    def _configure_cmake(self):
+        cmake = CMake(self)
+        cmake.definitions['IMGUI_DIR'] = os.path.join(self.source_folder, self._imgui_dir)
+        cmake.definitions['SFML_DIR'] = os.path.join(self.deps_cpp_info['sfml'].lib_paths[0], 'cmake', 'SFML')
```
eliasdaler: Need to set `SFML_STATIC_LIBRARIES` if self.options.shared is `False`.

conanfile.py
```diff
@@ -0,0 +1,111 @@
...
+      Whether or not to use -fPIC option while building
+
+    * imconfig: None or String
+      Use 'None' if you want this library to provide default
+      ImGui's imconfig.h file. Otherwise, provide a path to the
+      imconfig.h file you want to include in the build.
+
+    * imconfig_install_folder: None or String
+      Use 'None' if you want this build to include imconfig.h file,
+      otherwise use anything else.
```
eliasdaler: "use anything else" - better say "path to where your custom imconfig header will be installed"

eliasdaler: Also, there should be a way to not install custom imconfig.h (user might do it in their own CMake/conan recipe)

conanfile.py
```diff
@@ -0,0 +1,111 @@
...
+    exports = 'LICENSE.md'
+    _imgui_dir = 'imgui'
+
+    def source(self):
+        git = Git(folder=self._imgui_dir)
+        git.clone('https://github.com/ocornut/imgui.git', str(self.options.imgui_revision))
+
+    def configure(self):
+        imconfig = self.options.imconfig
+        if imconfig and not os.path.isfile(str(imconfig)):
```
eliasdaler: This makes me think... can we handle relative paths somehow? E.g. if I have `~/imgui-sfml/custom.h`, I want to run 
```sh
conan . create imgui-sfml/testing -o imconfig=custom.h
``` 

or at least 

```sh
conan . create imgui-sfml/testing -o imconfig=./custom.h
```

conanfile.py
eliasdaler: Actually, this raises the question - is there any way to SFML Conan package maintainers to specify that SFML_STATIC_LIBRARIES need to be defined when trying to link with static SFML libs?

conanfile.py
barsoosayque: I just tried that in the repository folder, and it worked. 
It also worked outside of the repository folder. So, relative paths work.

conanfile.py
barsoosayque: > Need to set `SFML_STATIC_LIBRARIES`

I didn't find any `SFML_STATIC_LIBRARIES` variables in SFML CMake nor in ImGui-SFML CMake. Is it something internal within CMake itself ?

Also, setting `self.options['sfml'].shared = ...` should resolve anything related to sfml and shared/static. Because otherwise it will be some kind of workaround.

> is there any way to SFML Conan package maintainers to specify that SFML_STATIC_LIBRARIES need to be defined when trying to link with static SFML libs?

I believe, they [already doing](https://github.com/bincrafters/conan-sfml/blob/testing/2.5.1/conanfile.py#L140) something like that.

eliasdaler: 'SFML_STATIC' is a preprocessor define. 'SFML_STATIC_LIBRARIES' is a CMake define and tells CMake that it needs to search for static SFML libraries.
Without it, I can't build ImGui-SFML as static lib, because it fails to find static SFML libs. See this:
https://github.com/SFML/SFML/blob/ff011dc51dc47373b0d49b087f8ac9d3d2bc8b06/cmake/SFMLConfig.cmake.in#L19

As an alternative (and I think it'll be better for ImGui-SFML) - let's set (SFML_STATIC_LIBRARIES as ON in ImGui's SFML's CMake is BUILD_SHARED_LIBS is set)

```cmake
if (BUILD_SHARED_LIBS)
    set(SFML_STATIC_LIBRARIES CACHE INTERNAL ON)
endif()
```

^ with this in ImGui-SFML's cmake, we won't need to set it in Conan recipe.

conanfile.py
eliasdaler: Strange, it didn't work for me. I think it depends where you build. If the current folder when you run "configure" is the same as your current folder, then yes, it will work.
But if you are building somewhere else, it won't. I'll test it again in a few hours.

conanfile.py
barsoosayque: Ah, I see. That's why I couldn't build it. 

As for the solution, I think it's it's more of a general build issue, so it's right to set this define in CMake in case of somebody building the library manually !

conanfile.py
barsoosayque: I will also try to verify this behavior in some corner cases. Because it seems very tricky.

conanfile.py
barsoosayque: > "path to where your custom imconfig header will be installed"

It's a bit confusing, actually, since:

> 24 | # If you set IMGUI_SFML_CONFIG_INSTALL_DIR, ImGui-SFML won't install your custom config, because
> 25 | # you might want to do it yourself

What does this variable do ? I think I just got it wrong.

conanfile.py
barsoosayque: Alright. I did find the error and fixed it, but the file won't get installed. I also might fix that in conan, though, I just use CMake install -- this means that plain build won't include specified file. 

Is this ok ?

conanfile.py
eliasdaler: Indeed. I think it's worth doing it in this PR.

conanfile.py
eliasdaler: Okay, the comment should say "otherwise specify the directory where your imconfig will be installed"

Plus, it's not just "imconfig.h" - it's "custom user config". I'll explain it all in other "conversation" here.

conanfile.py
eliasdaler: Okay, so here's how custom imconfig should be used.
If IMGUI_SFML_CONFIG_NAME is set, then IMGUI_SFML_CONFIG_DIR should be set to where this header can be found.
If IMGUI_SFML_CONFIG_INSTALL_DIR is not set, then this custom config will be installed with ImGui-SFML, otherwise it won't be installed, but it will use IMGUI_SFML_CONFIG_INSTALL_DIR, so when you link to ImGi-SFML CMake target, you'll get IMGUI_SFML_CONFIG_INSTALL_DIR in your include directories list.

So, imagine this. You have "~/imgui-sfml/my-config.h".
If you set this:

```
* IMGUI_SFML_USE_DEFAULT_CONFIG = OFF
* IMGUI_SFML_USER_CONFIG_DIR = ~/imgui-sfml
* IMGUI_SFML_USER_CONFIG_NAME = "my-config.h"
* IMGUI_SFML_CONFIG_INSTALL_DIR=~/my-custom-imgui-sfml/include
```

Then you'll get this:

1) During ImGui-SFML build, "~/imgui-sfml/my-config.h" will be used for building, but it won't be installed
2) If you link to ImGui-SFML, you'll get  "~/my-custom-imgui-sfml/include" to your include directories, because that's where "myconfig.h" will be when you install everything.

This is **very** complicated case, but I hope it's clear now.

eliasdaler:
Please merge master into your branch - this commit fixes some issues: https://github.com/eliasdaler/imgui-sfml/commit/b07b48f0ba4118ba4d830dee279a9ad2a596543b

conanfile.py
eliasdaler: So, here are the cases which I would like to see working (taking example further)
Suppose that pwd=~/imgui-sfml.

1) ```conan . create imgui-sfml/testing -o imconfig=my-config.h``` - this will us set:

```
* IMGUI_SFML_USE_DEFAULT_CONFIG = OFF
* IMGUI_SFML_USER_CONFIG_DIR = ~/imgui-sfml
* IMGUI_SFML_USER_CONFIG_NAME = "my-config.h"
```

```conan . create imgui-sfml/testing -o imconfig=./my-config.h``` should do the same

2) ```conan . create imgui-sfml/testing -o imconfig=my-config.h -o imconfig_install_folder=~/my-custom-imgui-sfml/include``` - this will us set:

```
* IMGUI_SFML_USE_DEFAULT_CONFIG = OFF
* IMGUI_SFML_USER_CONFIG_DIR = ~/imgui-sfml
* IMGUI_SFML_USER_CONFIG_NAME = "my-config.h"
* IMGUI_SFML_CONFIG_INSTALL_DIR= ~/my-custom-imgui-sfml/include
```

3) ```conan . create imgui-sfml/testing -o imconfig=~/custom-config/my-config.h -o imconfig_install_folder=~/my-custom-imgui-sfml/include``` - this will set:

```
* IMGUI_SFML_USE_DEFAULT_CONFIG = OFF
* IMGUI_SFML_USER_CONFIG_DIR = ~/custom-config
* IMGUI_SFML_USER_CONFIG_NAME = "my-config.h"
* IMGUI_SFML_CONFIG_INSTALL_DIR= ~/my-custom-imgui-sfml/include
```

eliasdaler:
Also I've released v2.0.2 just now, so you can rebase or merge it into your branch.

barsoosayque:
Merged v2.0.2 and fixed `imconfig` option. Hopefully, it works now !

eliasdaler:
Sorry, have been busy lately. Will try to look at it this week.

eliasdaler:
Hello, incredibly sorry I didn't look at it sooner- been having busy days at work... I'm on vacation now, so hopefully I'll find some time in the coming two weeks.

eliasdaler:
Merged in https://github.com/eliasdaler/imgui-sfml/commit/85f5d9a38c5d77979c1aea55c502b15ff9900d2d - had to squash commits for history to look cleaner.
Thanks for your work and patience.

eliasdaler:
Use [bincrafters recipe](https://bintray.com/bincrafters/public-conan/imgui-sfml%3Abincrafters/2.1%3Astable)
Unfortunately, it's not updated to the latest version yet, there are some unresolved problems here: https://github.com/bincrafters/community/pull/1368

PR #90, opened 2019-06-11
Added KeySuper update (fixes the inability to use shortcuts on macOS)

Melix19:
ImGUI uses the io.KeySuper instead of Ctrl on macOS for shortcuts, but there isn't an implementation on how KeySuper is setted, so doing any shortcut does nothing. This PR fixes that.


eliasdaler:
Looks good, thanks!

PR #91, opened 2019-06-14
Fix double new line input in InputTextMultiline (macOS)

Melix19:
Pressing enter in a InputTextMultiline on macOS produces two new lines instead of one. After some research I found that imgui already inputs a '\n' when pressing enter, but at the same time all the input chars taken from the sfml event are inserted in the InputText causing the double new line. This PR fixes this by excluding manually the new line char (IDK if this is the best way to do it, but this fixes the problem for me).
The strange fact is that this issue does not happen on Windows.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -395,13 +395,15 @@ void Update(const sf::Vector2i& mousePos, const sf::Vector2f& displaySize,
         }
     }
 
-    // Update Ctrl, Shift, Alt state
+    // Update Ctrl, Shift, Alt, Super state
     io.KeyCtrl = io.KeysDown[sf::Keyboard::LControl] ||
                  io.KeysDown[sf::Keyboard::RControl];
     io.KeyAlt =
         io.KeysDown[sf::Keyboard::LAlt] || io.KeysDown[sf::Keyboard::RAlt];
     io.KeyShift =
         io.KeysDown[sf::Keyboard::LShift] || io.KeysDown[sf::Keyboard::RShift];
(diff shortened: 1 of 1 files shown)


eliasdaler:
Also doesn't happen on Linux... Can you please take a look why ImGui doesn't instert "Enter" at least on Windows? I sadly don't have time to this for the next 1-2 days.

Melix19:
On Windows pressing enter produces a carriage return, instead on macOS it produces a line feed. The difference is that the carriage return sets the cursor to the beginning of the line, the line feed is basically a ‘\n’.

Melix19:
Little update: I tried both glfw and sdl2 to see if their text input events handle the enter char in the same way. The answer is no. Both doesn't produce any events on enter press. So, [as explained here](https://en.sfml-dev.org/forums/index.php?topic=16099.0), this is a peculiarity of SFML to just pass what the OS generates to SFML.
Now, I think that the best way to fix this is to filter this chars (\n, \r, \t, \b), as suggested in the topic linked above, and let imgui handle them as currently does.

Melix19:
Ok, so to recap the changes I have made:

- I have removed the char filter under 0x10000 (it's no more needed after [this commit](https://github.com/ocornut/imgui/commit/ef13d95466f27b4a955d6f95da90822001ace25f))
- I have removed the static cast to ImWchar (it's no more needed after the commit linked above)
- I have added a filter to \n (line feed), \r (carriage return), \t (tab) and \b (backspace) which are all handled inside ImGui

Melix19:
I have closed this pr because I used the wrong base commit to make changes and I had problems to fix it.
Opened #93.

PR #92, opened 2019-06-24
Add horizontal mouse wheel support

Melix19:
Now when using a mouse with a horizontal scroll wheel, the scroll is actually horizontal and not vertical.


eliasdaler:
Looks good. Please add curly braces after if/else conditions and I'll merge. :)

Melix19:
Done. I've also fixed the shift + vertical wheel that was not working because sfml recognizes directly as horizontal wheel scroll instead imgui handles it manually using the vertical scroll value.

eliasdaler:
Thanks, merged.

PR #93, opened 2019-06-24
Filter some characters that imgui already handles

Melix19:
This is what this pr changes:
- Removed the char filter under 0x10000 (it's no more needed after [this commit](https://github.com/ocornut/imgui/commit/ef13d95466f27b4a955d6f95da90822001ace25f))
- Removed the cast to ImWchar (it's no more needed after the commit linked above)
- Added a filter to \n (line feed), \r (carriage return), \t (tab) and \b (backspace) which are all handled inside ImGui (fixes the double linefeed bug in macOS #91)

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -314,10 +314,11 @@ void ProcessEvent(const sf::Event& event) {
                     (event.type == sf::Event::KeyPressed);
                 break;
             case sf::Event::TextEntered:
-                if (event.text.unicode > 0 && event.text.unicode < 0x10000) {
-                    io.AddInputCharacter(
-                        static_cast<ImWchar>(event.text.unicode));
+                // Don't handle the event for unprintable characters
+                if (event.text.unicode < ' ' || event.text.unicode == 127) {
+                    break;
                 }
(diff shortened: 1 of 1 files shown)


eliasdaler:
From ImGui's changelog:
> - Examples/Backends: Don't filter characters under 0x10000 before calling io.AddInputCharacter(),
  the filtering is done in io.AddInputCharacter() itself. This is in prevision for fuller Unicode
  support. (#2538, #2541)

So maybe we should add comparison `> 0x10000`?

eliasdaler:
https://github.com/spurious/SDL-mirror/blob/17af4584cb28cdb3c2feba17e7d989a806007d9f/src/events/SDL_keyboard.c#L794

Let's use this code to be close to what SDL does. You might even want to insert the same comment.

Melix19:
Done.

eliasdaler:
Looking at it, this comment feels wrong in context.
Let's say "Don't handle the event for unprintable characters"

Melix19:
Done.

eliasdaler:
Okay, merged, thanks again. :)

PR #95, opened 2019-07-19
Correct error message regarding IMGUI_DIR

AMS21:
Just notices while installing that the error message complains about IMGUI_ROOT not being set while the script actually checks for IMGUI_DIR.

Cheers and keep up the good work!


eliasdaler:
Thanks!

PR #99, opened 2019-08-04
Add option to use static SFML libraries to CMakeLists

salianifo:
I was trying to use CMake to compile ImGui-SFML but it was complaining that it couldn't find it, and I deduced that it was because my SFML libraries are built as static libs. Added this to make it work and thought I'd share.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ endif()
 
 option(IMGUI_SFML_BUILD_EXAMPLES "Build ImGui_SFML examples" OFF)
 option(IMGUI_SFML_FIND_SFML "Use find_package to find SFML" ON)
+option(IMGUI_SFML_USE_STATIC_LIBS "Use static SFML librares" ON)
 
 # If you want to use your own user config when compiling ImGui, please set the following variables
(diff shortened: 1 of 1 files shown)


eliasdaler:
Hello

Here's a thought. Maybe ImGui-SFML should look for static SFML automatically if you build ImGui-SFML as static lib? Building ImGui-SFML as shared lib, but linking to SFML statically can lead to some very bad results, and we'd want to protect against it.

So, I'm not really in favor of "use static SFML with ImGui-SFML" option just because ImGui-SFML can't find SFML if it's built statically only - it won't help user understant that they need to set `SFML_STATIC_LIBRARIES` to `ON`.

salianifo:
I didn't see the BUILD_SHARED_LIBS option when I was running CMake, so I didn't know that option was there. After your comment, I read through the CMakeLists file and saw that at the bottom, but it doesn't look like it's defined anywhere?

eliasdaler:
It doesn't need to be defined for

```if (NOT BUILD_SHARED_LIBS)``` to work ;)

eliasdaler:
The patch looks good, merged.

PR #105, opened 2019-09-08
[ADD] support for downloading ImGui with git

NoeMurr:
This commit add the support for automatically downloading ImGui in a temp directory.
The git_clone cmake function  has been taken from [this repo](https://github.com/tschuchortdev/cmake_git_clone.git).

I've also changed the FindImGui.cmake file because of a little bug: 

if you look for imgui.h after having installed imgui-sfml on your machine, you will find the imgui.h file in the /usr/local directory no matter what the value of IMGUI_DIR is set to. 
But that is not the directory where the sources of ImGui are. 

If you look for the imgui.cpp file you will find the correct directory.


eliasdaler:
Hello.
Thanks for the patch, however I think the better approach would be one of the things:
1) Use ImGui as subrepo of ImGui-SFML
2) Use [ExternalData](https://cmake.org/cmake/help/git-master/module/ExternalData.html)

ExternalData already implements cloning git repos, so I see no reason to repeat this, considering it's probably done a lot safer and better in CMake.

> if you look for imgui.h after having installed imgui-sfml on your machine, you will find the imgui.h file in the /usr/local directory no matter what the value of IMGUI_DIR is set to.
But that is not the directory where the sources of ImGui are.
Strange, I've tested it, and it should be installed in the place where imgui-sfml is put into. Try setting ``CMAKE_INSTALL_PREFIX`` to some directory and see where imgui headers get installed.

Also, the value of ```IMGUI_DIR``` is used for finding ImGui, it doesn't affect where ImGui headers get installed to. Both ImGui and ImGui-SFML headers are installed to  ```CMAKE_INSTALL_INCLUDEDIR``. Also see [this](https://github.com/eliasdaler/imgui-sfml/issues/80#issuecomment-491864232) discussion.

I'll double-check where imgui.h gets installed, though.

So, can't accept it for now, sorry. I'll be glad to accept the patch which either includes ImGui as a subrepo (see #81) or downloads it via ExternalData (which should be configurable and OFF by default).

PR #123, opened 2020-06-20
Rename SFML::ShowTestWindow to SFML::ShowDemoWindow

sabidib:
According to the imgui changelog v1.53 of imgui deprecated `SFML::ShowTestWindow`. It kept the function around for a while until v1.75 which removed the reference. This means that compiling with `IMGUI_SFML_BUILD_EXAMPLES=ON` fails with the most recent of SFML.

I've adjusted the main.cpp to use `SFML::ShowDemoWindow`

### DIFF
--- a/examples/main.cpp
+++ b/examples/main.cpp
@@ -28,7 +28,7 @@ int main()
 
         ImGui::SFML::Update(window, deltaClock.restart());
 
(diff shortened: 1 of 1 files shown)


eliasdaler:
Thanks!

Issue #121, opened 2020-05-25
Include demo sources


pfirsich:
Hello,
It would be nice if there was a switch to include `IMGUI_DEMO_SOURCES` (i.e. `imgui_demo.cpp`) in the build (or to include it in the build). I am not very experienced with ImGui and find `ShowDemoWindow` very useful.

eliasdaler:
Agreed. Will add `IMGUI_SFML_IMGUI_DEMO` option to CMake build

eliasdaler:
Well, I was very busy, but thankfully @dubgron provided a PR, so now you can set `IMGUI_SFML_IMGUI_DEMO` to `ON` to build imgui_demo.cpp.

PR #124, opened 2020-06-25
Added IMGUI_SFML_IMGUI_DEMO option to CMake build

dubgron:
Enabling it will result in compiling imgui_demo.cpp file, that contains e.g. ImGui::ShowDemoWindow function, which is useful for new users.

Resolves: #121

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,7 @@ find_package(OpenGL REQUIRED)
 add_library(ImGui-SFML
   imgui-SFML.cpp
   ${IMGUI_SOURCES}
+  $<$<BOOL:${IMGUI_SFML_IMGUI_DEMO}>:${IMGUI_DEMO_SOURCES}>
 )
 
 # Add pretty alias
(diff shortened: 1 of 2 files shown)


eliasdaler:
Hello
I think you should add it as `option` in main CMakeLists.txt file and in FindImGui.cmake have `IMGUI_DEMO_SOURCE` variable defined. And then:

```cmake
if (IMGUI_SFML_IMGUI_DEMO)
    list(APPEND IMGUI_SOURCES ${IMGUI_SFML_IMGUI_DEMO})
endif()

add_library(ImGui-SFML
  imgui-SFML.cpp
  ${IMGUI_SOURCES}
)

...
```

dubgron:
Hi,
I thought about doing it that way, but in my opinion using CMake's generator expressions results in nicer code. But if you prefer it your way I made a new commit.

eliasdaler:
Please also add this in the root CMakeLists:

```diff
option(IMGUI_SFML_FIND_SFML "Use find_package to find SFML" ON)
+ option(IMGUI_SFML_IMGUI_DEMO "Build imgui_demo.cpp" OFF)
```

The `option` is a good way of documenting that some variables can be set by user, to differentiate them from internal script variables. It's a bad practice to have non-cache variables being set by user externally.

dubgron:
Good point.

eliasdaler:
Thanks, merged!

PR #137, opened 2020-11-09
enable ImGui_SFML to be built when SFML was being added as subdirectory

lulitao1997:
right now, when including SFML and ImGui_SFML as dependencies and building them from source
like following
```
add_subdirectory(third_party/SFML)
SET(IMGUI_DIR "...")
add_subdirectory(third_party/imgui-sfml)

target_link_libraries(main sfml-system sfml-window sfml-graphics ImGui-SFML::ImGui-SFML)
```

cmake will fail to call find_package(SFML), to fix that we
should not call `find_package` when there's already sfml-graphics etc
targets exist (this is the situation where we include SFML as
subdirectory instead of using system SFML).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,6 @@ if(POLICY CMP0077)
 endif()
 
 option(IMGUI_SFML_BUILD_EXAMPLES "Build ImGui_SFML examples" OFF)
-option(IMGUI_SFML_FIND_SFML "Use find_package to find SFML" ON)
 option(IMGUI_SFML_IMGUI_DEMO "Build imgui_demo.cpp" OFF)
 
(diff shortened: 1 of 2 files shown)


eliasdaler:
That's exactly what why `IMGUI_SFML_FIND_SFML` flag was added. If you're adding SFML via `add_subdirectory`, you should just set `IMGUI_SFML_FIND_SFML` to `FALSE`, as done [here](https://github.com/eliasdaler/imgui-sfml-fetchcontent/blob/master/dependencies/imgui-sfml/CMakeLists.txt#L22). If you do it, `find_package` won't be called.

I don't think that checking for SFML's targets is a better solution.

lulitao1997:
Ah, I thought it didn't work, turns out i need to reconfigure my cmake on my machine, sorry for bothering

closing this PR

PR #144, opened 2020-12-09
Update FindImGui.cmake to include imgui_tables.cpp

mxgrey:
It seems the source file `imgui_tables.cpp` has been added upstream to imgui, so the cmake build fails with a lot of linking errors if that source file isn't included in this list.


eliasdaler:
I'll merge this when ImGui v 1.80 gets released.

eliasdaler:
Actually, the way GitHub actions are written now imply that we're tracking ImGui's master, so I've added this while I was fixing Github Actions failing. Thanks for the notice, this change is already in master (and sorry that I didn't merge this to show your authorship)

PR #150, opened 2021-01-23
Fixed inclusion of imgui_tables.cpp source file

fgallegosalido:
Hello! Thank you for this amazing library!

I added a fix to the inclusion of the source file `imgui_tables.cpp` inside the `cmake/FindImGui.cmake` file. As you may know, this source file is part of the new release of Dear ImGui v1.80, but it was missing in prior releases.

For that reason, ImGui-SFML was refusing to build for versions of Dear ImGui lower than v1.80, because that specific file was missing. I just confitionally included the file in the `IMGUI_SOURCES`, depending on the version of Dear ImGui.

If you think the condition should be different (maybe a specific commit or something else), feel free to modify whatever is necessary.


eliasdaler:
Looks good, thanks!

PR #152, opened 2021-02-06
Update to README.md to include imgui_tables.cpp

alexanderremm:
The latest version of imgui requires that the imgui_tables.cpp file be included in order to use the library. I've updated the README manual install instructions to include this change.


eliasdaler:
Looks good, thanks

PR #162, opened 2021-03-25
Fix 'bgColor' not being passed to 'ImGui::ImageButton'

SuperV1234:
Always compile with warnings enabled! :)


eliasdaler:
Thanks a lot! Yeah, I need to add some better warnings to the CI. :)

PR #163, opened 2021-03-27
Fix warning message C4100

utilForever:
- warning C4100: 'bgColor': unreferenced formal parameter


eliasdaler:
Hello
Thanks for the PR, but I'd prefer if we got rid of this warning by commenting-out argument's name (e.g. `const sf::Color& /* bgColor */`) instead of using Dear ImGui's `IM_UNUSED` macro, as it looks like it's meant to be used in Dear ImGui code only.

eliasdaler:
Hello
It doesn't seem like it's needed after https://github.com/eliasdaler/imgui-sfml/pull/162. Can you please test on the latest master? The C4100 should be gone by now - it was an error because `bgColor` was not passed into `ImGui::ImageButton`, but it's now passed, so it's not unreferenced.

utilForever:
Oh, I'm so sorry. I tested on the latest master and confirmed the build is passed. 😂

PR #176, opened 2021-07-14
Add version to CMake install

Srykah:
This PR allows users to provide a version to find_package, with a same major version compatibility.
Even though I changed the readme to reflect this change, providing a version is optional, so it doesn't break anything for current users 😉


eliasdaler:
Hello
Sorry, missed this PR. 
I think that major version should be left out of this PR, because I don't really do major/minor versions well (I break ABI from time  to time...), so it's not worth for users to really care about them (maybe I should follow ImGui's versioning and say that each release might break something in ABI)

PR #178, opened 2021-08-10
Support #define ImDrawIdx unsigned int in renderer via imconfig.h

Doy-lee:
In `imconfig.h` we can enable support for more than 64k of vertices in the draw list be enabling the hash define see: https://github.com/ocornut/imgui/blob/f99fe72c42bd5d578c2443409192b139b1c6623d/imconfig.h#L96

This can be supported in this framework by making sure GL knows the correct size of the index buffer when we go to render.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -842,7 +842,8 @@ void RenderDrawLists(ImDrawData* draw_data) {
                     // Bind texture, Draw
                     GLuint textureHandle = convertImTextureIDToGLTextureHandle(pcmd->TextureId);
                     glBindTexture(GL_TEXTURE_2D, textureHandle);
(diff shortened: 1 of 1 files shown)


eliasdaler:
Hello
This looks good, but I try to be closer to ImGui's OpenGL backend impl, so please just change it to this and I'll merge :)

```cpp
glDrawElements(GL_TRIANGLES, (GLsizei)pcmd->ElemCount, sizeof(ImDrawIdx) == 2 ? GL_UNSIGNED_SHORT : GL_UNSIGNED_INT, idx_buffer);
```

Doy-lee:
Commit has been updated :)

eliasdaler:
Thanks, merged.

PR #189, opened 2021-12-12
Use a map instead of a vector for window contexts

fgallegosalido:
Currently, the window contexts are stored in a `std::vector`. This forces us to loop through it in order to find a specific WindowContext by the system handle of the window.

Using a `std::map` allows us to just look for it using the system handle of the window as the key, and thus avoiding many loops. The ideal solution would be to use `std::unordered_set` and define a hash for the class `WindowContext`, but this requires at least C++11, which should be considered for future releases of ImGui-SFML, as SFML 3 will use C++17.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -20,7 +20,7 @@
 #include <cstddef> // offsetof, NULL, size_t
 #include <cstring> // memcpy
 
-#include <vector>
+#include <map>
 
 #ifdef ANDROID
 #ifdef USE_JNI
@@ -232,7 +232,8 @@ struct WindowContext {
     }
 };
 
-std::vector<WindowContext*> s_windowContexts;
+typedef std::map<sf::WindowHandle, WindowContext*> WindowContextsMap;
+WindowContextsMap s_windowContexts;
 WindowContext* s_currWindowCtx = NULL;
 
 } // end of anonymous namespace
@@ -254,9 +255,9 @@ void Init(sf::Window& window, const sf::Vector2f& displaySize, bool loadDefaultF
                                                    // GLuint.
 #endif
 
-    s_windowContexts.push_back(new WindowContext(&window));
+    s_currWindowCtx = new WindowContext(&window);
+    s_windowContexts.insert({window.getSystemHandle(), s_currWindowCtx});
 
-    s_currWindowCtx = s_windowContexts.back();
     ImGui::SetCurrentContext(s_currWindowCtx->imContext);
 
     ImGuiIO& io = ImGui::GetIO();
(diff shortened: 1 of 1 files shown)


imgui-SFML.cpp
```diff
@@ -529,45 +529,39 @@ void Render() {
...
-                    ImGui::SetCurrentContext(s_currWindowCtx->imContext);
-                    break;
-                }
-            }
+            s_currWindowCtx = s_windowContexts.begin()->second;
+            ImGui::SetCurrentContext(s_currWindowCtx->imContext);
         } else {
             // no alternatives...
             s_currWindowCtx = NULL;
+            ImGui::SetCurrentContext(NULL);
```
oprypin, line 554: It seems that this `SetCurrentContext(NULL)` is a change in behavior, previously there wasn't such an action.

eliasdaler: Not just `unordered_map`, but `std::unique_ptr` should be used as well, probably. :)

oprypin (approved):
I like this change, I think it should be merged as is, and then further improvements can be done with a smaller diff

oprypin (changes requested):
Nope nevermind, it introduces a crash on the multi-window example.

`c++ -I. -I../cimgui/imgui examples/multiple_windows/main.cpp imgui-SFML.cpp ../cimgui/imgui/{imgui_demo.cpp,imgui_widgets.cpp,imgui.cpp,imgui_draw.cpp,imgui_tables.cpp} -lsfml-graphics -lsfml-window -lsfml-system -lGL && ./a.out`

*close the Child Window*

```
a.out: imgui-SFML.cpp:533: void ImGui::SFML::Shutdown(const sf::Window&): Assertion `false && "Window wasn't inited properly: forgot to call ImGui::SFML::Init(window)?"' failed.
zsh: IOT instruction (core dumped)  ./a.out
```

imgui-SFML.cpp
```diff
@@ -529,45 +529,39 @@ void Render() {
...
+    // check if the current window's context is the one we want to remove
+    const bool isCurrentWindowCtx =
+        window.getSystemHandle() == s_currWindowCtx->window->getSystemHandle();
+
+    // remove window's context  
+    delete s_windowContexts[window.getSystemHandle()];
+    s_windowContexts.erase(window.getSystemHandle());
+
     // set current context to some window for convenience if needed
-    if (window.getSystemHandle() == s_currWindowCtx->window->getSystemHandle()) {
```
oprypin, line 533: At this point:
```c++
    printf("%d %d\n", window.getSystemHandle(), s_currWindowCtx->window->getSystemHandle());
```
on Linux this prints `0 0`.

That is, after a window is closed, its handle changes to 0.
So basically previously it kinda worked by accident, the condition is basically this: if the handle of the passed window is closed and the handle of the current window is closed, assume they're the same window.
Because the usage of `map` assumes the `SystemHandle` will stay constant and valid, the new logic is not equivalent and causes a crash.

oprypin:
Because I think this use of `map` is not viable, I keep `vector` but make most of the other suggested improvements in my PR https://github.com/eliasdaler/imgui-sfml/pull/202

eliasdaler:
Superseded by #202

PR #191, opened 2021-12-21
Propagate possible texture creation failure to Init

vittorioromeo:


eliasdaler:
Looks good, however afa3839 removed the examples for some reason.
Can you restore them, please? :)
(and looks like I need to add examples to CI if it didn't fail...)

eliasdaler:
Thanks, merged

PR #192, opened 2021-12-24
Image drawing sf::Sprite respects transformation

DNKpp:
A ``sf::Sprite`` drawn via Image function doesn't respect the transformaiton. That's what I wanted to fix.
This pull request is rather WIP, but this way its easier to talk about improvements and actual behaviour.

The sprite transformation works (proof below) but I had to apply some changes to the interface. This is not a big deal, because it should be easy to simply add my function  and call it from the previous two overloads, thus no breaking change necessary. But here is my second struggle. The positioning and size behaviour of the ``sf::RenderTexture``, ``sf::Texture`` and ``sf::Sprite`` now seem to differ quite a bit. I wasn't able to determine how a ``size`` param should change the appearance of such a sprite, as a sprite itself has a complete transformation. I decided to get rid of that overload and instead add a ``sf::Transform`` param, because this is much more flexible than the provided interface of a ``sf::Sprite``. With a ``sf::Transform`` its possible to perform transformations in different order; the ``sf::Sprite`` uses a fixed ordering.

Looking forward for any feedback!

image

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -701,9 +701,8 @@ void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Transfo
 {
     const sf::IntRect& textureRect = sprite.getTextureRect();
 
-    // sprite without or empty texture cannot be drawn
+    // sprite without texture on empty frame cannot be drawn
     if (!sprite.getTexture() ||
-        textureRect.width <= 0 || textureRect.height <= 0 ||
         size.x <= 0 || size.y <= 0) {
         return;
     }
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -1,5 +1,6 @@
 #include "imgui-SFML.h"
 #include <imgui.h>
+#include <imgui_internal.h>
 
 #include <SFML/Config.hpp>
 #include <SFML/Graphics/Color.hpp>
@@ -116,6 +117,7 @@ namespace {
 ImColor toImColor(sf::Color c);
 ImVec2 getTopLeftAbsolute(const sf::FloatRect& rect);
 ImVec2 getDownRightAbsolute(const sf::FloatRect& rect);
+void toImVec2Quad(const sf::FloatRect& rect, const sf::Transform& transform, ImVec2 quad[4]);
 
 ImTextureID convertGLTextureHandleToImTextureID(GLuint glTextureHandle);
 GLuint convertImTextureIDToGLTextureHandle(ImTextureID textureID);
@@ -686,29 +688,59 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
 /////////////// Image Overloads for sf::Sprite
 
 void Image(const sf::Sprite& sprite, const sf::Color& tintColor, const sf::Color& borderColor) {
-    sf::FloatRect bounds = sprite.getGlobalBounds();
-    Image(sprite, sf::Vector2f(bounds.width, bounds.height), tintColor, borderColor);
+	Image(sprite, sf::Transform::Identity, tintColor, borderColor);
 }
 
-void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Color& tintColor,
-           const sf::Color& borderColor) {
+void Image(const sf::Sprite& sprite, const sf::Transform& additionalTransform, const sf::Color& tintColor,
+    const sf::Color& borderColor)
+{
     const sf::Texture* texturePtr = sprite.getTexture();
     // sprite without texture cannot be drawn
     if (!texturePtr) {
         return;
     }
 
+    ImGuiWindow* window = GetCurrentWindow();
+    if (window->SkipItems)
+        return;
+
     const sf::Texture& texture = *texturePtr;
+    ImTextureID textureID = convertGLTextureHandleToImTextureID(texture.getNativeHandle());
     sf::Vector2f textureSize = static_cast<sf::Vector2f>(texture.getSize());
     const sf::IntRect& textureRect = sprite.getTextureRect();
     ImVec2 uv0(textureRect.left / textureSize.x, textureRect.top / textureSize.y);
     ImVec2 uv1((textureRect.left + textureRect.width) / textureSize.x,
                (textureRect.top + textureRect.height) / textureSize.y);
 
-    ImTextureID textureID = convertGLTextureHandleToImTextureID(texture.getNativeHandle());
+    const ImVec2 uvs[4] =
+    {
+        uv0,
+        ImVec2(uv1.x, uv0.y),
+        uv1,
+        ImVec2(uv0.x, uv1.y)
+    };
 
-    ImGui::Image(textureID, ImVec2(size.x, size.y), uv0, uv1, toImColor(tintColor),
-                 toImColor(borderColor));
+    sf::FloatRect spriteRect = sprite.getLocalBounds();
+    if (borderColor.a > 0)
+    {
+        spriteRect.left += 1;
+        spriteRect.top += 1;
+        spriteRect.width -= 2;
+        spriteRect.height -= 2;
+    }
+
+    const sf::Transform transform = additionalTransform * sprite.getTransform();
+    ImVec2 pos[4];
+    toImVec2Quad(spriteRect, transform, pos);
+
+    if (borderColor.a > 0)
+    {
+        ImVec2 borderPos[4];
+        toImVec2Quad(sprite.getLocalBounds(), transform, borderPos);
+        window->DrawList->AddQuad(borderPos[0], borderPos[1], borderPos[2], borderPos[3], toImColor(borderColor), 0.f);
+    }
+
+    window->DrawList->AddImageQuad(textureID, pos[0], pos[1], pos[2], pos[3], uvs[0], uvs[1], uvs[2], uvs[3], toImColor(tintColor));
 }
 
 /////////////// Image Button Overloads for sf::Texture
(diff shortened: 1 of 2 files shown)


DNKpp:
Well, I think I solved most of my struggles. I simply scaled the sprite to ``size`` after all other transformations have been applied. Additionally I improved the handling of the sprite, as it is now properly handled by ImGui (mostly copy-pasted from original ``ImGui::Image`` function).
One question is still unsolved. How to handle the border? Is it expected to apply transformations to the border rect or should it stay uneffected? Currently it is transformed, but I wonder if there ever is an actual usecase for this... (look at the picture above, its very far from prett) On the other handside this is the more difficult part. If one simply wants a border around the sprite frame, than it should be very easy to do so, without expecting the function to do it.

imgui-SFML.cpp
```diff
@@ -688,39 +688,61 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
-    const sf::Texture& texture = *texturePtr;
-    ImTextureID textureID = convertGLTextureHandleToImTextureID(texture.getNativeHandle());
-    sf::Vector2f textureSize = static_cast<sf::Vector2f>(texture.getSize());
-    const sf::IntRect& textureRect = sprite.getTextureRect();
+    ImRect itemBB(window->DC.CursorPos.x, window->DC.CursorPos.y,
+        window->DC.CursorPos.x + size.x, window->DC.CursorPos.y + size.y);
+    if (borderColor.a > 0)
+    {
+        itemBB.Max.x += 2;
+        itemBB.Max.y += 2;
```
eliasdaler, line 720: Why is this needed?

imgui-SFML.cpp
```diff
@@ -688,39 +688,61 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
+    }
+
+    ItemSize(itemBB);
+    if (!ItemAdd(itemBB, 0))
+        return;
+
+    // prepare uv coordinates
+    const sf::Vector2f textureSize = static_cast<sf::Vector2f>(sprite.getTexture()->getSize());
```
eliasdaler, line 728: Texture can be NULL here. Please add a check and just "return" if it's the case

imgui-SFML.cpp
```diff
@@ -688,39 +688,61 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
 }
 
-void Image(const sf::Sprite& sprite, const sf::Transform& additionalTransform, const sf::Color& tintColor,
-    const sf::Color& borderColor)
+void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Color& tintColor,
+    const sf::Color& borderColor) {
+	Image(sprite, size, sf::Transform::Identity, tintColor, borderColor);
+}
+
+void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Transform& additionalTransform,
```
eliasdaler, line 699: Change "additionalTransform" to just "transform", please

imgui-SFML.cpp
```diff
@@ -688,39 +688,61 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
                (textureRect.top + textureRect.height) / textureSize.y);
 
-    const ImVec2 uvs[4] =
-    {
-        uv0,
-        ImVec2(uv1.x, uv0.y),
-        uv1,
-        ImVec2(uv0.x, uv1.y)
-    };
+    sf::Transform transform = additionalTransform * sprite.getTransform();
```
eliasdaler, line 733: This can be named "finalTransform"

imgui-SFML.cpp
```diff
@@ -688,39 +688,61 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
+	sf::FloatRect spriteRect = sprite.getLocalBounds();
+	const sf::FloatRect bounding = transform.transformRect(spriteRect);
+
+    // applies the transformations which are expected as a child of the parent window
+    sf::Transform childTransform;
+    childTransform.translate(itemBB.Min.x, itemBB.Min.y).
+        scale(size.x / bounding.width, size.y / bounding.height).
```
eliasdaler, line 740: Possible division by zero?

imgui-SFML.cpp
```diff
@@ -1,5 +1,6 @@
 #include "imgui-SFML.h"
 #include <imgui.h>
+#include <imgui_internal.h>
```
eliasdaler, line 3: Why is this needed?

imgui-SFML.cpp
```diff
@@ -116,6 +117,7 @@ namespace {
 ImColor toImColor(sf::Color c);
 ImVec2 getTopLeftAbsolute(const sf::FloatRect& rect);
 ImVec2 getDownRightAbsolute(const sf::FloatRect& rect);
+void toImVec2Quad(const sf::FloatRect& rect, const sf::Transform& transform, ImVec2 quad[4]);
```
eliasdaler, line 120: This function should return quad instead of accepting it as a parameter

imgui-SFML.cpp
```diff
@@ -686,29 +688,85 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
+	Image(sprite, size, sf::Transform::Identity, tintColor, borderColor);
+}
+
+void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Transform& additionalTransform,
+    const sf::Color& tintColor, const sf::Color& borderColor)
+{
+    const sf::IntRect& textureRect = sprite.getTextureRect();
+
+    // sprite without texture on empty frame cannot be drawn
+    if (!sprite.getTexture() ||
```
eliasdaler, line 705: This check should be done before getting textureRect

imgui-SFML.cpp
```diff
@@ -686,29 +688,85 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
+        translate(-bounding.getPosition());
+    transform = childTransform * transform;
+    
+
+    if (borderColor.a > 0)
+    {
+        spriteRect.left += 1;
+        spriteRect.top += 1;
+        spriteRect.width -= 2;
+        spriteRect.height -= 2;
```
eliasdaler, line 750: Are the borders always 1px wide?

imgui-SFML.cpp
```diff
@@ -686,29 +688,85 @@ void Image(const sf::RenderTexture& texture, const sf::Vector2f& size, const sf:
...
+
+    ImRect itemBB(window->DC.CursorPos.x, window->DC.CursorPos.y,
+        window->DC.CursorPos.x + size.x, window->DC.CursorPos.y + size.y);
+    if (borderColor.a > 0)
+    {
+        itemBB.Max.x += 2;
+        itemBB.Max.y += 2;
+    }
+
+    ItemSize(itemBB);
```
eliasdaler, line 723: Prefix all Dear ImGui functions with "ImGui" namespace please. Otherwise it's not clear that we're using ImGui's functions

eliasdaler (commented):
Wow, this is pretty complicated...
I'd say that it's better to leave overload which called ImGui::Image as it's much simpler to understand and a lot more lightweight for most uses.

Yes, we'd have some duplication in "extended" version, but at least only one of the functions would break when ImGui::Image changes its algorithm. Plus, using <imgui_internal.h> makes e a bit uneasy, because it doesn't have any API compatibility promises. But I guess nothing can be done about it...

eliasdaler:
"One question is still unsolved. How to handle the border? Is it expected to apply transformations to the border rect or should it stay uneffected? Currently it is transformed, but I wonder if there ever is an actual usecase for this... (look at the picture above, its very far from pretty) On the other handside this is the more difficult part. If one simply wants a border around the sprite frame, than it should be very easy to do so, without expecting the function to do it."

Does Dear ImGui have any widgets that can be rotated? If it does, please see how it transforms the borders.
If it doesn't - please try to create an implementation of this complex overload outside of ImGui-SFML's code. If it's possible, then maybe we should provide this as a code snippet/example. I really don't want to support a code which uses imgui_external and does so many complex things not related to the actual backend implementation, to be honest. :(

imgui-SFML.cpp
DNKpp: Similar behaviour as in original ImGui::Image. Simply ensures, that actual drawn image is exactly the requested size, when a border shall be drawn.

imgui-SFML.cpp
DNKpp: It is already checked in the first condition of the function...

imgui-SFML.cpp
DNKpp: Not check necessary, as the texture rect is a value member of ``sf::Sprite``: https://github.com/SFML/SFML/blob/master/include/SFML/Graphics/Sprite.hpp

imgui-SFML.cpp
DNKpp: Well, that's exactly what the original ``ImGui::Image`` function assumes.

DNKpp:
> Does Dear ImGui have any widgets that can be rotated? If it does, please see how it transforms the borders. If it doesn't - please try to create an implementation of this complex overload outside of ImGui-SFML's code. If it's possible, then maybe we should provide this as a code snippet/example.
> 
Actually don't know. But the I think the best solution would be to simply draw the border independent to the transformation.
> I really don't want to support a code which uses imgui_external and does so many complex things not related to the actual backend implementation, to be honest. :(
What do you expect? Transformation handling isn't just straight forward. Creating bridge code between two independant domains, neither. Additionally you require to keep things c++03 conform.
As I have to emulate the internals of ``ImGui::Image``, inclusion of ``imgui_internal.h`` is simply necessary. Otherwise such functionas like ``ItemSize`` and ``ItemAdd`` will not be available.

imgui-SFML.cpp
DNKpp: Due to necessary calls to ``ItemSize`` and ``ItemAdd``

imgui-SFML.cpp
DNKpp: If you tell me how to return a c array from a function, without allocating heap mem or declaring it static, I'll do that.

DNKpp:
I've refactored the code accordingly to your comments.
I also decided to simply draw the border without the sprite transformation. This looks like this:
image

imgui-SFML.cpp
DNKpp: Yes, added an early return

eigenl:
Perhaps this could be merged then? Will need this functionality personally.

eliasdaler:
After some thought, I'm closing this.
The complexity and future maintenance cost is not worth it. 

Everything here can be done without making it a part of ImGui-SFML.

PR #193, opened 2022-02-03
Wrapping updateMouseCursor() behind a dirty flag, like other example backends. Fixes the window resize cursor not being shown.

lucasz93:
As described in https://en.sfml-dev.org/forums/index.php?topic=27180.0 and https://github.com/eliasdaler/imgui-sfml/issues/119.

`updateMouseCursor` was being spammed every tick which wasn't letting Windows temporarily override the IMGUI's cursor with the resize cursor. Other backends ([like the Win32 one](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_win32.cpp#L358)) wrap their equivalent of `updateMouseCursor` behind a dirty flag.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -173,6 +173,7 @@ struct WindowContext {
     bool windowHasFocus;
     bool mouseMoved;
     bool mousePressed[3];
+    ImGuiMouseCursor lastCursor;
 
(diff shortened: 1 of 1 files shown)


eliasdaler:
Hello.
Thank you for the PR! Sorry it took me so long to get to it.

PR #197, opened 2022-03-12
Update to use AddKeyEvent, supporting the full key range from v1.87

### LINKED ISSUE #174: State of modifier keys is wrong in certain scenarios
I'm not sure if the root cause is within SFML itself or imgui-SFML, but you can observe the following behavior:

1. Create a combobox with enough items so that the dropdown list has a scrollbar. Or simply use ImGui::ShowDemoWindow().
2. Open the combobox, now you can use the mousewheel to scroll through the droplist.
3. Press `Ctrl` once.
4. Now scrolling through the droplist via mousewheel no longer works.
5. Press any "normal" key (like a letter).
6. Scrolling works again.

The problem is within `ProcessEvent()`, which sets the keystate with `io.KeyCtrl = event.key.control;`. However, `event.key.control` seems to contain the wrong state in this scenario, probably because in this scenario `Ctrl` is treated as normal key and not as modifier key in combination with another key like e.g. `Ctrl+C`.
Pressing a "normal" key afterwards resets `io.KeyCtrl` again, so scrolling works again.
As mentioned, I'm not sure if this isn't an issue within SFML itself.

This scenario is particularly annoying to me because I have an application where I use `Ctrl+Mousewheel` to zoom in/out. So it is commonly used and afterwards the comboboxes don't work correctly anymore because the of this.

A rather crude fix that works for me is to replace `io.KeyCtrl = event.key.control;` with:

```
if( (event.key.code == sf::Keyboard::LControl) || (event.key.code == sf::Keyboard::RControl) )
{
  io.KeyCtrl = (event.type == sf::Event::KeyPressed);
}
else
{
  io.KeyCtrl = event.key.control;
}
```

I have only checked with the `Ctrl` key, but I assume other modifiers like `Alt` or `Shift` suffer from the same problem.

### LINKED ISSUE #196: Missing support for new events in ImGui v1.87
[Dear ImGui v1.87](https://github.com/ocornut/imgui/releases/tag/v1.87) has expanded its support for keyboard and gamepad buttons, and replaced the interface for backends to report the keys. The old way keeps working OK but is deprecated and has big limitations, thus already making "imgui-demo" run with some missing functionality when it's run through "imgui-sfml".

---

Code such as `ImGui::IsKeyDown(ImGuiKey_Q)` is supposed to work for backends that are up to date with ImGui v1.87, but imgui-sfml never reports to ImGui that that key was pressed, because in the past the set of available keys was limited:

https://github.com/eliasdaler/imgui-sfml/blob/4129d276d45845581b6ba99ede50db6f761e5089/imgui-SFML.cpp#L284-L291

---

Here you can observe the missing events:
One can just run the default "imgui-demo" and in the current state of imgui-sfml it is missing a lot of information in the "Keyboard, Gamepad & Navigation State".

current behavior | expected behavior
-|-
https://user-images.githubusercontent.com/371383/158024702-86a606c2-aea7-4718-a494-6e5ec8b06510.mp4  | https://user-images.githubusercontent.com/371383/158024705-83840c9a-26ad-41ad-86a5-073bf89f51fd.mp4

---

How I was running this:

```diff
diff --git a/examples/minimal/main.cpp b/examples/minimal/main.cpp
index 2c19abd08..e0561a94b 100644
--- a/examples/minimal/main.cpp
+++ b/examples/minimal/main.cpp
@@ -12,6 +12,10 @@ int main() {
     window.setFramerateLimit(60);
     ImGui::SFML::Init(window);
 
+    ImGuiIO& io = ImGui::GetIO();
+    io.ConfigFlags |= ImGuiConfigFlags_NavEnableKeyboard;
+    io.ConfigFlags |= ImGuiConfigFlags_NavEnableGamepad;
+
     sf::CircleShape shape(100.f);
     shape.setFillColor(sf::Color::Green);

```

`c++ -I. -I./imgui examples/minimal/main.cpp imgui-SFML.cpp ./imgui/{imgui_demo.cpp,imgui_widgets.cpp,imgui.cpp,imgui_draw.cpp,imgui_tables.cpp} -lsfml-graphics -lsfml-window -lsfml-system -lGL && ./a.out`

And here's how I was running with ImGui's own SDL backend, to be able to compare:
`c++ -I. -Ibackends -I/usr/include/SDL2 -lSDL2 -lGL examples/example_sdl_opengl3/main.cpp backends/imgui_impl_opengl3.cpp backends/imgui_impl_sdl.cpp imgui_demo.cpp imgui_widgets.cpp imgui.cpp imgui_draw.cpp imgui_tables.cpp && ./a.out`

oprypin:
[Dear ImGui v1.87](https://github.com/ocornut/imgui/releases/tag/v1.87) has expanded its support for keyboard and gamepad buttons, and replaced the interface for backends to report the keys. The old way keeps working OK but is deprecated and has big limitations, thus already making "imgui-demo" run with some missing functionality when it's run through "imgui-sfml".

This change implements the new approach. So, (almost) all the keyboard keys and possible gamepad events will be properly reported to ImGui and possible to query.
Fixes #196.
Fixes #174.

I implemented this by following what [changes the SDL backend has undergone](https://github.com/ocornut/imgui/compare/fe646ea59131fb9bd792de8843173b7e6f6b9e51~..c906c65cac6860e7e77649100cdf8fbf1bb201a0#diff-6cf0dc51e38fab2be9ad7660aacc471a658522e02ce53dede1ca6c3b2a177e42) and trying to make equivalent changes here.

---

Here you can observe the missing events:
One can just run the default "imgui-demo" and in the current state of imgui-sfml it is missing a lot of information in the "Keyboard, Gamepad & Navigation State".
After this change it matches the SDL backend's behavior.

before | after
-|-
https://user-images.githubusercontent.com/371383/158024702-86a606c2-aea7-4718-a494-6e5ec8b06510.mp4  | https://user-images.githubusercontent.com/371383/158024705-83840c9a-26ad-41ad-86a5-073bf89f51fd.mp4

---

This is backwards-compatible in the sense that I'm pretty sure all user code will continue working as is.
In particular, both of these keep working the same:
```
ImGui::IsKeyDown(sf::Keyboard::Z) == ImGui::IsKeyDown(ImGui::GetKeyIndex(ImGuiKey_Z))
```

In the table below, "before" means without this change and "after" means with this change.

&nbsp; | ImGui 1.86 before | ImGui 1.87 before | ImGui 1.87 after
-:|:-:|:-:|:-:
`ImGui::IsKeyDown(sf::Keyboard::Z)` | Works | Works | Works
`ImGui::IsKeyDown(ImGui::GetKeyIndex(ImGuiKey_Z))` | Works | Works | Works
`ImGui::IsKeyDown(ImGuiKey_Z)` | Always false | Works | Works

But <kbd>Z</kbd> was one of the keys that was possible to make ImGui aware of.
<kbd>Q</kbd> was **not** one of such keys, hence:

&nbsp; | ImGui 1.86 before | ImGui 1.87 before | ImGui 1.87 after
-:|:-:|:-:|:-:
`ImGui::IsKeyDown(sf::Keyboard::Q)` | Works | Works | Works
`ImGui::IsKeyDown(ImGui::GetKeyIndex(ImGuiKey_Q))` | Invalid code | Always false | Works
`ImGui::IsKeyDown(ImGuiKey_Q)` | Invalid code | Always false | Works

---

From a different point of view, this is **not** backwards-compatible because it requires ImGui 1.87 to compile.
If the tables above had a column "ImGui 1.86 after", that would all be "imgui-sfml can't compile".

---

I also change the deadzone for sticks from 5% to 15%, because that seems much closer to what the SDL backend does.

---

Additions to the API are also up for discussion.

```diff
 IMGUI_SFML_API void SetJoytickDPadThreshold(float threshold);
 IMGUI_SFML_API void SetJoytickLStickThreshold(float threshold);
+IMGUI_SFML_API void SetJoytickRStickThreshold(float threshold);
+IMGUI_SFML_API void SetJoytickLTriggerThreshold(float threshold);
+IMGUI_SFML_API void SetJoytickRTriggerThreshold(float threshold);
```

In particular:
Should we keep the typo "Joytick" consistent?
Should we perhaps merge the L + R pairs of these functions into one function?


eliasdaler:
Hello.
Wow, this is a fantastic PR, thanks a lot!

I'll take a look at it soon as I'm currently very busy with real life stuff... 

As for the "Joytick" - feel free to rename it to "Joystick". I have no idea how no one noticed that one before. API/ABI break is okay in our case (since Dear ImGui and SFML don't have make any ABI promises)

> Should we perhaps merge the L + R pairs of these functions into one function?
Two functions are better than one, imo.

oprypin:
> Which OSes did you test this on?

Only on [Arch Linux (KDE, Xorg, NVIDIA)](https://pryp.in/specs/2022-03-09.html)

> how no one noticed that one before

I definitely [noticed](https://github.com/oprypin/crystal-imgui-sfml/blame/f30de7df64527eb447fbea658a317555b63f0b95/src/imgui-sfml.cr#L101), but didn't say anything

eliasdaler:
> Only on [Arch Linux (KDE, Xorg, NVIDIA)](https://pryp.in/specs/2022-03-09.html)

Okay, I see. I'll test it on Ubuntu and Windows 10 a bit later!

Please rebase your branch onto the latest master, I'll approve the Gitlab Actions runs (CI was broken)

eliasdaler:
Please sync with master again - there are some merge conflicts now. :)

eliasdaler:
Tested on Ubuntu 20.04 and Windows 10 - works fine.
The one failing job is CI's issues, not caused by this branch.
Merging... thank you!

PR #202, opened 2022-04-01
Refactor: reduce use of pointers and array indices

oprypin:
+ The `Texture` and `Cursor` members don't need to be pointers, they are empty by default.
+ For `s_windowContexts` use `unique_ptr` instead.

+ Refactor Shutdown function to not need to explicitly skip the current window.
+ SetCurrentContext to null when no windows remain.


eliasdaler:
Thanks a lot! That's a really well-done PR :)

PR #205, opened 2022-04-23
Replaced deprecated sf::Keyboard::Return with sf::Keyboard::Enter

Bambo-Borris:
Whilst working on a project using the latest snapshot of SFML master branch (where the SFML v3 commits are going) I noticed that the `sf::Keyboard::Return` was no longer marked deprecated but had now been removed which caused `imgui-sfml.cpp` to fail to compile since it still relied on this enum entry. Replaced the entry with the preferred `sf::Keyboard::Enter`

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -350,7 +350,7 @@ ImGuiKey keycodeToImGuiKey(sf::Keyboard::Key code) {
         return ImGuiKey_Backspace;
     case sf::Keyboard::Space:
         return ImGuiKey_Space;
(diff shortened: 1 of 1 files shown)


oprypin (approved):
Indeed this is deprecated. And I checked that no other deprecated keys are present, so it's good to merge

eliasdaler:
Thank you, merged.

Bambo-Borris:
> Thank you, merged.

Glad to have given even a small change to a repo that has helped me many times in projects.

PR #213, opened 2022-06-15
Viewports

nocheacosador:
Hi!
I've just implemented ImGui multi-viewport feature for SFML. I haven't tested it with multiple ImGui contexts but with one context it seems to work. I am planning on developing it further, especially if you would consider merging it with the master branch or creating a new viewports branch in your repository. Please, don't hesitate to give me any feedback, I'm eager to contribute and learn :) 
Hope to hear from you soon.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -24,6 +24,14 @@
 #include <memory>
 #include <vector>
 
+
+// For multi-viewport support enable/disable
+#define VIEWPORTS_ENABLE
+
+#if defined(VIEWPORTS_ENABLE) && defined(_WIN32)
+#include <Windows.h>
+#endif
+
 #ifdef ANDROID
 #ifdef USE_JNI
 
@@ -176,7 +184,7 @@ struct TriggerInfo {
 };
 
 struct WindowContext {
-    const sf::Window* window;
+    sf::Window* const window;
     ImGuiContext* imContext;
 
     sf::Texture fontTexture; // internal font atlas which is used if user doesn't set a custom
@@ -207,12 +215,24 @@ struct WindowContext {
 #endif
 #endif
 
+#ifdef VIEWPORTS_ENABLE
+    const bool imContextOwner; // Context owner/main viewport
+    const bool isRenderWindow;
+
     WindowContext(const WindowContext&) = delete; // non construction-copyable
     WindowContext& operator=(const WindowContext&) = delete; // non copyable
 
-    WindowContext(const sf::Window* w) {
-        window = w;
-        imContext = ImGui::CreateContext();
+    WindowContext(sf::RenderWindow* w, ImGuiContext* context = nullptr) 
+        : WindowContext(w, context, true) { }
+
+    WindowContext(sf::Window* w, ImGuiContext* context = nullptr, bool isRenderWindow = false) 
+        : window(w), imContextOwner(context == nullptr), isRenderWindow(isRenderWindow)
+    {
+        if (context) {
+            imContext = context;
+        } else {
+            imContext = ::ImGui::CreateContext();
+        }
 
         windowHasFocus = window->hasFocus();
         mouseMoved = false;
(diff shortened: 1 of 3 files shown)


nocheacosador:
Ah. Also, My implementation will only work on Windows because ImGui requires monitor info and as far as I know SFML doesn't provide any API for that, so I used Windows API for that, in the future I would be able to implement it for Ubuntu. As for MacOS it would be difficult cause I don't have Mac machine accessible to me to test the implementation.

sam20908:
Thanks for your work on this! My project uses imgui-sfml and we've tried to use imgui's docking feature with no success. Will this PR allow us to use docking?

nocheacosador:
> Thanks for your work on this! My project uses imgui-sfml and we've tried to use imgui's docking feature with no success. Will this PR allow us to use docking?

Well, as far as I know docking should work straight out of the box, even without this PR. You just need to use docking branch of ImGui and enable docking with setting `ImGuiConfigFlags_EnableDocking` flag in `io.ConfigFlags`.

eliasdaler:
Hello.
Thanks for the PR. It's pretty complex and I can't promise that I'll be able to review it in depth soon (maybe in 2-3 weeks? Lots of stuff going on in real life, sorry...)

Is this one based on SDL backend implementation? I think that SFML and SDL implementations should be pretty close.

imgui-SFML.cpp
```diff
@@ -207,15 +216,28 @@ struct WindowContext {
 #endif
 #endif
 
+#ifdef VIEWPORTS_ENABLE
```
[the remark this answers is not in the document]
eliasdaler: Ah, I see, the `WindowContext` constructor might become ambiguous because of default arguments... You shouldn't make them default them to avoid ambiguity then, in my opinion.

imgui-SFML.cpp
```diff
@@ -666,17 +765,178 @@ void Update(sf::Window& window, sf::RenderTarget& target, sf::Time dt) {
...
+        const bool windowNoInput = (viewport->Flags & ImGuiViewportFlags_NoInputs) != 0;
+
+        if (wc->mouseHovered && !windowNoInput) mouseViewportID = viewport->ID;
+    }
+
+    if (io.BackendFlags & ImGuiBackendFlags_HasMouseHoveredViewport)
+        io.AddMouseViewportEvent(mouseViewportID);
+
+#else
     if (s_currWindowCtx->windowHasFocus) {
```
eliasdaler, line 940: Why doesn't this code need to be called when viewports are on?

imgui-SFML.cpp
```diff
@@ -1403,4 +1676,183 @@ void updateMouseCursor(sf::Window& window) {
     }
 }
 
+
+#ifdef VIEWPORTS_ENABLE
+
+void SFML_CreateWindow(ImGuiViewport* viewport) {
+#if SFML_VERSION_MAJOR >= 3
+    sf::RenderWindow* window = new sf::RenderWindow(sf::VideoMode({(unsigned)viewport->Size.x,
```
eliasdaler, line 1684: Should be `unsigned int`, not just `unsigned`

imgui-SFML.cpp
```diff
@@ -1403,4 +1676,183 @@ void updateMouseCursor(sf::Window& window) {
...
+#endif
+    window->setVisible(false);
+    WindowContext* data = IM_NEW(WindowContext)(window, ImGui::GetCurrentContext(), true);
+    data->windowHasFocus = true;
+    viewport->PlatformUserData = data;
+    viewport->PlatformHandle = window;
+}
+
+void SFML_DestroyWindow(ImGuiViewport* viewport) {
+    if (WindowContext* wc = (WindowContext*)viewport->PlatformUserData) {
```
eliasdaler, line 1700: Is wc ever `nullptr` here?

imgui-SFML.cpp
```diff
@@ -1403,4 +1676,183 @@ void updateMouseCursor(sf::Window& window) {
...
+    WindowContext* wc = (WindowContext*)viewport->PlatformUserData;
+    wc->window->setVisible(true);
+}
+
+void SFML_SetWindowPos(ImGuiViewport* viewport, ImVec2 pos) {
+    WindowContext* wc = (WindowContext*)viewport->PlatformUserData;
+    wc->window->setPosition(pos);
+}
+
+ImVec2 SFML_GetWindowPos(ImGuiViewport* viewport) {
```
eliasdaler, line 1720: Why doesn't sf::Window::getPosition not work here for all platforms?

imgui-SFML.cpp
```diff
@@ -1403,4 +1676,183 @@ void updateMouseCursor(sf::Window& window) {
...
+    WindowContext* wc = (WindowContext*)viewport->PlatformUserData;
+
+#ifdef _WIN32    
+    RECT clientAreaRect;
+    HWND hwnd = wc->window->getSystemHandle();
+    GetClientRect(hwnd, &clientAreaRect);
+    MapWindowPoints(hwnd, NULL, (LPPOINT)&clientAreaRect, 2);
+    return { (float)clientAreaRect.left, (float)clientAreaRect.top };
+#else
+    if (wc->imContextOwner) return wc->window->getPosition() + sf::Vector2i(0, 24);
```
eliasdaler, line 1730: What is this (0, 24) offset?

imgui-SFML.cpp
```diff
@@ -1403,4 +1676,183 @@ void updateMouseCursor(sf::Window& window) {
...
+    WindowContext* wc = (WindowContext*)viewport->PlatformUserData;
+    wc->window->setTitle(str);
+}
+
+void SFML_UpdateWindow(ImGuiViewport* viewport) {
+}
+
+void SFML_RenderWindow(ImGuiViewport* viewport, void*) {
+    WindowContext* wc = (WindowContext*)viewport->PlatformUserData;
+    if (!wc->imContextOwner) {
```
eliasdaler, line 1770: Easier to read:

```cpp
if (wc->imContextOwner) {
   return;
}

... // rest of the code
```

eliasdaler (changes requested):

eliasdaler:
@oprypin, @vittorioromeo 
Would appreciate if you take a look at this if viewports/docking are something you're interested in. :)

Also, it seems to me like it would be great to separate imgui-sfml into several files now. Any thoughts?
Looks like imgui-SFML.cpp is growing way too big to handle easily...

imgui-SFML.cpp
nocheacosador: On Windows SFML `sf::Window::getPosition()` returns top left corner position in screen coords including title bar. ImGui requires position of client area (area where stuff is drawn), therefore if value returned by `sf::Window::getPosition()` is used, main viewport becomes offsetted in reference to `io.MousePos` by title bar height.

imgui-SFML.cpp
nocheacosador: This was an early attempt to cancel out that title bar offset in the main viewport. I will delete this.

imgui-SFML.cpp
nocheacosador: Because it is called while iterating through viewports. In ImGui main window is also considered a viewport and it is the first item in `ImGuiPlatformIO::Viewports` vector.

imgui-SFML.cpp
nocheacosador: No.

imgui-SFML.cpp
```diff
@@ -666,17 +765,178 @@ void Update(sf::Window& window, sf::RenderTarget& target, sf::Time dt) {
...
+    
     io.DisplaySize = ImVec2(displaySize.x, displaySize.y);
     io.DeltaTime = dt.asSeconds();
 
+#ifdef VIEWPORTS_ENABLE
+    if (io.ConfigFlags & ImGuiConfigFlags_ViewportsEnable) {
+        SFML_UpdateMonitors();
+
+        for (auto& viewport : ImGui::GetPlatformIO().Viewports) {
+            WindowContext* wc = (WindowContext*)viewport->PlatformUserData;
```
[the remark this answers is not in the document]
nocheacosador: `ImGui::SFML::ProcessEvent` handles only main window events. Events for viewports' windows are not handled by it. Maybe, to avoid copy-pasting, `ImGui::SFML::ProcessEvent` could call internal function, for example `ProcessEvent(const Event&, WindowContext*)`, with `s_currWindowCtx`, and in this loop it would be called with corresponding viewport's window context?

sam20908:
@oprypin looks like code for Linux has been added does that interest you?

MirrasHue:
Amazing, man!

I was even wondering about using the official backend (GLFW probably) to have this feature, but I prefer this one much better, so that I can stick with SFML as well.

dpieve:
> Well, as far as I know docking should work straight out of the box, even without this PR. You just need to use docking branch of ImGui and enable docking with setting `ImGuiConfigFlags_EnableDocking` flag in `io.ConfigFlags`.

I don't think docking should work straight out of the box. I tried to render the SFML inside a docking window, but no success either. Maybe I'm missing something? 

It crashes and stops at some point, saying that begins/ends don't match.

The following code doesn't crash, but I couldn't make it render the circle inside a docking window. For this example, 2 docking windows will be displayed but no circle.

Any suggestion how I could fix it?

```cpp
#include "imgui-SFML.h"
#include "imgui.h"
#include "SFML/Graphics.hpp"
int main() {
    sf::RenderWindow window(sf::VideoMode(800, 800), "ImGui Docking SFML");

    ImGui::SFML::Init(window);

    ImGuiIO& io = ImGui::GetIO();
    io.ConfigFlags = ImGuiConfigFlags_DockingEnable;

    sf::CircleShape shape(200.f, 100);
    shape.setFillColor(sf::Color(200, 70, 5)); 

    sf::Clock deltaClock;

    while (window.isOpen())
    {
        sf::Event event;
        while (window.pollEvent(event))
        {
            ImGui::SFML::ProcessEvent(event);
            if (event.type == sf::Event::Closed)
                window.close();
        }

        ImGui::SFML::Update(window, deltaClock.restart());

        ImGui::DockSpaceOverViewport(ImGui::GetMainViewport());

        ImGui::Begin("Dock test", NULL, ImGuiConfigFlags_DockingEnable);
		    ImGui::ShowDemoWindow();
        ImGui::End();

        window.clear(sf::Color(18, 33, 43));
        window.draw(shape);
        ImGui::SFML::Render(window);
        window.display();

    }
    ImGui::SFML::Shutdown();
    return 0;
}
```

anthony-sv:
Why is this still a pull request? SHOULD HAVE BEEN MERGED already.

ctag-fh-kiel:
Is this pull request still active, am interested if it works with mac and could test it...

eliasdaler:
Vitorrio, please don’t add me to reviews anymore. 

I haven’t used SFML for many years and don’t have capacity to maintain/review ImGui-SFML stuff anymore.

PR #217, opened 2022-09-18
Replace sf::Uint8 with std::uint8_t

pifopi:
Hello !

Just a small update because sfml recently replaced sf::Uint8 with std::uint8_t
https://github.com/SFML/SFML/commit/e2528de20a58d56c13d22fe64fd55d701ba3c714

The CI fails with mingw errors but it is not linked to my change (only submitting an empty CL to trigger the CI leads to the exact same issue).

Thanks a lot for your work !

### DIFF
--- a/imconfig-SFML.h
+++ b/imconfig-SFML.h
@@ -3,6 +3,8 @@
 #include <SFML/Graphics/Color.hpp>
 #include <SFML/System/Vector2.hpp>
 
+#include <cstdint>
+
 #include "imgui-SFML_export.h"
 
 #define IM_VEC2_CLASS_EXTRA           \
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
For the record this is a change in the in-development major update to SFML, not in the 2.x series. If we still want to support both SFML 2 and 3 without separate branches and we're okay with using C++11, we can safely switch to the standard fixed width aliases which SFML 3 is already using.

ChrisThrasher (approved):

eliasdaler:
Just to be clear - this change still works with SFML 2.5, right?
(also mingw CI is failing for some reason not related to this PR, I don't have time to fix it right now, unfortunately)

ChrisThrasher:
> Just to be clear - this change still works with SFML 2.5, right?

I think so? I'm not sure how it would break.

oprypin:
I mean, it could be slightly painful if these types appeared anywhere at the boundary of the API of imgui-sfml, but they don't. Yeah it's safe. And CI passes too, so of course it has to be fine.

pifopi:
Yes, it's fine with SFML 2.5 ^^

eliasdaler:
Okay, thanks, everyone! Merged.

PR #222, opened 2022-10-02
making sure imgui works well after resizing

hulakdar:
right now imgui windows are clipping to old size and mouse position is not tracked well after resizing


eliasdaler:
Looks good, thanks.

Issue #206, opened 2022-05-08
Race condition with Window.hasFocus()


danieljpetersen:
imgui-sfml relies on the false assumption that Window.hasFocus() will report true immediately after creation. Unfortunately, there is a short period of time after window creation where Window.hasFocus() is not yet true. This is the case for both macOS, Windows, and likely Linux though not confirmed there.

This results in an intermittent bug where input such as mouse clicks will be skipped / not recorded.

This was an issue reported by l04 on the SFML Discord. You can find a (spread out making it somewhat hard to follow) conversation about the issue starting here, 
https://discord.com/channels/175298431294636032/175298431294636032/972715469310148658

Quoting l04 here 

> In the constructor of WindowContext, at line 217 of imgui-SFML.cpp, the statement windowHasFocus = window->hasFocus() is called. 

> However, the window might not have the focus yet, so it returns false.

> Then in the function ImGui::SFML::ProcessEvent, the events LostFocus and GainedFocus are checked.

> However, the window might have got the focus before calling the function, so it will not register the GainedFocus event.
> In this case WindowContext is stuck with windowHasFocus = false until you manually generate the GainedFocus event.
> If I replace line 217 of imgui-SFML.cpp with windowHasFocus = true, I can't reproduce the error anymore. Maybe it needs a better way to handle the initial focus?

A simple fix would be to set the windowHasFocus variable to true immediately on init rather than checking with the function call, though this is admittedly an ugly fix and I'm not sure that this is applicable in all cases. For example in the case of multiple windows.

oprypin:
I didn't fully understand the symptoms. 

This is the important part:

> there is a short period of time after window creation where Window.hasFocus() is not yet true



> This results in an intermittent bug where input such as mouse clicks will be skipped / not recorded.

Not recorded for how long, or until what happens? How haven't others noticed this?

danieljpetersen:
Apologies, I should have been more clear.

> I didn't fully understand the symptoms.

When this bug is triggered, most sf::Events will not be  forwarded to imgui until the user does something to cause an sf::Event::LostFocus, and then an sf::Event::GainedFocus. 

In other words, when this bug is triggered the user needs to do something like click outside of the window and then click back onto the window, or alt tab in order to get events to propagate from sfml to imgui.

The following events are not forwarded to imgui when the bug is triggered:
> sf::Event::MouseMoved
sf::Event::MouseButtonPressed
sf::Event::MouseButtonReleased
[...]

Issue #88, opened 2019-06-02
Alt-Tabbing keeps the `io.KeysDown[sf::Keyboard::Tab]` to true


Epholys:
Hello!

I found that, when alt-tabbing to another window, the `io.KeysDown[sf::Keyboard::Tab]` stays  `true`. It's because, in `ProcessEvent()`, the events are not processed when the window loses focus. As a consequence, for example, it is impossible to open/close `CollapsingHeader` when going back to the SFML window after alt-tabbing, as "Alt" is considered pressed.

An easy hack is possible, but I think a proper fix could be developed. I don't know how you would prefer to implement it, so I open this issue instead of a pull request.

Thank you for this project!

Cheers

eliasdaler:
Hello. Yes, I'll try to fix it, but let's talk about what should happen.
I think that when window loses the focus, all members of `io.KeyDown` should become `false`. Then, when window has focus restored, I'll poll the state of all relevant keys, so that I get "now" state of keyboard right.

I think I'll also poll keyboard state at startup, so if you start your program with some keyboard key already pressed, ImGui-SFML will register that.

Epholys:
I think that's a good idea! But there's maybe some other inputs to take care of, like joysticks or controllers.

And there's something I just thought off and that seems strange: when alt-tabbing back to the SFML window, there is the event "GainedFocus". So the "KeyReleased" of "Alt" should be registered and "KeysDown" set to false. But that's not the case. Maybe there's something not so obvious with the SFML event polling and/or the OS platform. I'm using Ubuntu with Gnome. Maybe it will have an influence on the fix.

eliasdaler:
Can you please test how ImGui's example OpenGL or SDL binding behave?
Looking at them, I don't see how they handle this.
Maybe they handle all keydown/keyup events even if you don't have window active?

eliasdaler:
I think that I can make ImGui-SFML handle all events regardless of window being focused.
This way, it'll be up to user to decide if they want this or not - e.g. just not call `ImGui::SFML::ProcessEvent` when window has lost its focus.

Epholys:
I can't test it right now, we'll have to wait a little. I quickly looked at them, I think you're right, they do not care if the window is focused or not... but it's maybe because there are just very simple examples.

I think that's a good idea, it'll probably remove this bug and probably other related unknown ones. And it could be a little more future-proof with the docking-and-more features of a next imgui release.

pkdawson:
[...]

PR #227, opened 2022-11-30
Don't care about focus when processing events / updating

### LINKED ISSUE #212: Events not handled when apps start without focus
Hello,

When I start the app without focus (e.g. when I click anywhere in my screen before the window to show) I cannot drag window, click on buttons and all imgui events aren't available.

I can reproduce the error with the basic example given in the readme.md. I am on windows and the error happen with both g++ 11.2.0 and clang++ 13.0.0

I don't know if it's the right place to tell that or if I must say it on the general imgui repository.

Besides that everything works perfectly.

eliasdaler:
Fixes #212, #206, #88 
Now the user has to not call Update/ProcessEvent on lost focus manually.
Might break some things (e.g. trigger some asserts if the user didn't call NewFrame/Render on focus lost).

@oprypin @ChrisThrasher, please review

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -182,7 +182,6 @@ struct WindowContext {
     sf::Texture fontTexture; // internal font atlas which is used if user doesn't set a custom
                              // sf::Texture.
 
-    bool windowHasFocus;
     bool mouseMoved;
     bool mousePressed[3];
     ImGuiMouseCursor lastCursor;
@@ -214,7 +213,6 @@ struct WindowContext {
         window = w;
         imContext = ImGui::CreateContext();
 
-        windowHasFocus = window->hasFocus();
         mouseMoved = false;
         for (int i = 0; i < 3; ++i) {
             mousePressed[i] = false;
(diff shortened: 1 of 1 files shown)


eliasdaler:
How does SDL backend manage it? Does it keep track of which window is active?

And can’t it be left to the user? E.g. making sure to only call ProcessEvent of the currently focused window, but not others?

JonnyPtn:
SDL backend uses a combination of SDL_GetMouseFocus() + SDL_HINT_MOUSE_FOCUS_CLICKTHROUGH to check whether a window is hovered without focus, but this isn't something SFML supports so I don't think there's anything you can do about it

To me this change seems sensible - windows don't get events anyway when they aren't focussed so adding extra checks in your code is redundant and just adds risk of issues (as reported)

JonnyPtn (approved):

oprypin:
I don't think it's OK that clicking on an unrelated application will send clicks to an imgui-sfml application that happens to be hidden behind it.

JonnyPtn:
> I don't think it's OK that clicking on an unrelated application will send clicks to an imgui-sfml application that happens to be hidden behind it.

How does this happen? SFML only receives events if the window has focus

oprypin:
Just run it and see for yourself if you don't believe

imgui-SFML.cpp
```diff
@@ -680,21 +666,18 @@ void Update(const sf::Vector2i& mousePos, const sf::Vector2f& displaySize, sf::T
...
+    if (io.WantSetMousePos) {
+        sf::Vector2i newMousePos(static_cast<int>(io.MousePos.x), static_cast<int>(io.MousePos.y));
+        sf::Mouse::setPosition(newMousePos);
+    } else {
+        io.MousePos = ImVec2(static_cast<float>(mousePos.x), static_cast<float>(mousePos.y));
+    }
+    for (unsigned int i = 0; i < 3; i++) {
+        io.MouseDown[i] = s_currWindowCtx->touchDown[i] || sf::Touch::isDown(i) ||
+                          s_currWindowCtx->mousePressed[i] ||
+                          sf::Mouse::isButtonPressed((sf::Mouse::Button)i);
```
JonnyPtn, line 678: @oprypin  I see what you're saying now - it's because this directly queries the mouse state here which bypasses the event queue.

Ideally think this should use events instead?

JonnyPtn (changes requested):

imgui-SFML.cpp
eliasdaler: Yeah, it looks like this is what causes problems, thanks.
I guess I should really rework how ProcessEvent/Update/Render etc. work and not use any direct queries since they don't respect window focus. 

Of course, maybe if I check `s_currWindowCtx.window->hasFocus()` it will solve the problem, but it feels somewhat hacky.

eliasdaler:
I guess hover events happen because we're changing `io.MouseDown` - we can just not do that when window has no focus (same with `io.MouseDown`, gamepad etc.)
But it looks like it lands us at square one. 🤔

JonnyPtn:
Yeah I don’t think you can do anything about hovering on partially visible windows unless you write the platform implementations for it yourself. 

If you just change to use events then users would still be able to implement this themselves if desired

Issue #228, opened 2023-01-25
Minor Breaking Change in SFML2.6


Jim-Marsden:
First I would like to say that I will gladly do a PR for this, I just don't know how you'd want this done as far as branching.

sf::Keyboard::Quote and sf::Keyboard::Tilde were both changed recently in SFML 2.6. I know SFML is working to releasing SFML 2.6. 

In ImGui::keycodeToImGuiKey
sf::Keyboard::Quote -> sf::Keyboard::Apostrophe
sf::Keyboard::Tilde   -> sf::Keyboard::Grave

oprypin:
It's not a **breaking** change in 2.6, only in 3.0. It is best to keep the existing code working until any release happens

https://github.com/SFML/SFML/blob/2.6.x/include/SFML/Window/Keyboard.hpp#L169
https://github.com/SFML/SFML/blob/8a307df494d1a8934384d8861cad3a1b9dbaaaba/include/SFML/Window/Keyboard.hpp#L169

eXpl0it3r:
It's not breaking, but they are deprecated, so might make sense to update them. 🙂

oprypin:
Sorry that my message is so decisive. I don't actually participate in this project. Mainly wanted to clarify.

Jim-Marsden:
> Sorry that my message is so decisive. I don't actually participate in this project. Mainly wanted to clarify.

No, you were exactly correct. I was on the wrong branch when I thought it was removed. So quickly closed everything.

As it is deprecated in 2.6 it might be an issue for later, but as far as my issue goes, I was wrong.

PR #231, opened 2023-01-30
Update imgui-SFML.cpp

pifopi:
Hello !

Quote and Tilde were deleted in the latest SFML
https://github.com/SFML/SFML/commit/72cd8781bbe35009c0a1660de4daf54394ca24fa

Thanks in advance !

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -354,7 +354,7 @@ ImGuiKey keycodeToImGuiKey(sf::Keyboard::Key code) {
         return ImGuiKey_Enter;
     case sf::Keyboard::Escape:
         return ImGuiKey_Escape;
-    case sf::Keyboard::Quote:
+    case sf::Keyboard::Apostrophe:
         return ImGuiKey_Apostrophe;
     case sf::Keyboard::Comma:
         return ImGuiKey_Comma;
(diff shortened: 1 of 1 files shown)


eliasdaler:
I don't follow SFML's master, I use the latest version
See #228. After SFML 2.6 gets released, I'll accept this commit.
Otherwise it will force everyone to use the latest SFML if they want to get new ImGui-SFML updates.

eliasdaler:
However, this makes me a bit uncomfortable that people who use ImGui-SFML won't be able to use "bleeding edge" SFML, so maybe I should make some branch which allows you to use SFML's "master" with ImGui-SFML?

(I won't be able to maintain it myself, though, someone will need to do it)

pifopi:
I can probably take care of this if you wish. Changes shouldn't be too massive anyway ! I checked and sadly only one of those two changes is compatible with sfml 2.5 The other isn't as far as I could see

eliasdaler:
Hmm... I think the better solution is to do something like this:

```cpp
#ifndef IMGUI_SFML_MASTER
    #define IMGUI_SFML_KEY_APOSTROPHE sf::Keyboard::Quote
    #define IMGUI_SFML_KEY_GRAVE sf::Keyboard::Tilde
#else
    #define IMGUI_SFML_KEY_APOSTROPHE sf::Keyboard::Apostrophe
    #define IMGUI_SFML_KEY_GRAVE sf::Keyboard::Grave
#endif

pifopi:
I deleted my fork yesterday so i'm creating an other PR with that fix

pifopi:
https://github.com/eliasdaler/imgui-sfml/pull/232

PR #232, opened 2023-01-31
fix for SFML master integration without breaking for non master builds

pifopi:


imgui-SFML.cpp
```diff
@@ -25,6 +25,14 @@
 #include <memory>
 #include <vector>
 
+#ifdef SFML_VERSION_MAJOR >= 3
```
oprypin, line 28:
```suggestion
#if SFML_VERSION_MAJOR >= 3
```

I think it needs to be written like this.
currently there is an error.

```
/Users/runner/work/imgui-sfml/imgui-sfml/imgui-sfml/imgui-SFML.cpp:28:27: warning: extra tokens at end of #ifdef directive [-Wextra-tokens]
#ifdef SFML_VERSION_MAJOR >= 3
                          ^
                          //
/Users/runner/work/imgui-sfml/imgui-sfml/imgui-sfml/imgui-SFML.cpp:365:10: error: no member named 'Apostrophe' in 
```

pifopi: thanks indeed !

eliasdaler:
Thank you, merged!

PR #235, opened 2023-02-20
Fix some warnings

vittorioromeo:
Fixes various warnings, including: "unhandled values in switch", "implicit conversion from signed to unsigned".

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -231,7 +231,7 @@ struct WindowContext {
         lastCursor = ImGuiMouseCursor_COUNT;
 
         joystickId = getConnectedJoystickId();
-        for (int i = 0; i < sf::Joystick::ButtonCount; ++i) {
+        for (int i = 0; i < static_cast<int>(sf::Joystick::ButtonCount); ++i) {
             joystickMapping[i] = ImGuiKey_None;
         }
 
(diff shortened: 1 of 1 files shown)


imgui-SFML.cpp
```diff
@@ -231,7 +231,7 @@ struct WindowContext {
         lastCursor = ImGuiMouseCursor_COUNT;
 
         joystickId = getConnectedJoystickId();
-        for (int i = 0; i < sf::Joystick::ButtonCount; ++i) {
+        for (int i = 0; i < static_cast<int>(sf::Joystick::ButtonCount); ++i) {
```
ChrisThrasher, line 234: Looks like an SFML/SFML bug we ought to fix. We shouldn't use a signed int for a count like this. But that's outside the scope of this PR.

imgui-SFML.cpp
```diff
@@ -1164,7 +1168,7 @@ void RenderDrawLists(ImDrawData* draw_data) {
     }
 
     ImGuiIO& io = ImGui::GetIO();
-    assert(io.Fonts->TexID != (ImTextureID)NULL); // You forgot to create and set font texture
+    assert(io.Fonts->TexID != (ImTextureID) nullptr); // You forgot to create and set font texture
```
ChrisThrasher, line 1171: Do we even need to do that cast to `ImTextureID`? Maybe that's out of scope of this PR though.

eliasdaler: I think this cast was needed to calm down some other compiler warnings :)

eliasdaler (approved):

PR #242, opened 2023-07-02
Require SFML 2.6

ChrisThrasher:
Even though ImGui-SFML doesn't technically need SFML 2.6, we ought to update this library in lockstep with SFML for the sake of giving developers a consistent experience between this library and SFML. I tried to find all the places we're referring to 2.5 and upgrade them.

What else do we need to do before we can release v2.6?


CMakeLists.txt
```diff
@@ -94,8 +79,6 @@ add_library(ImGui-SFML::ImGui-SFML ALIAS ImGui-SFML)
 target_link_libraries(ImGui-SFML
   PUBLIC
     sfml-graphics
-    sfml-system
-    sfml-window
```
eliasdaler, line 98: Did you remove it because sfml-graphics links to them as well? In my opinion, it's better to link to these modules explicitly as ImGui-SFML uses functions from them

eliasdaler:
> What else do we need to do before we can release v2.6?

I guess it's good enough. Other things can be done/fixed in v2.6.x and breaking changes in 3.x.x, I guess

CMakeLists.txt
ChrisThrasher: It's just wasted code if you ask me but I'll put it back. SFML itself relies on these transitive dependencies so there's strong precedent for doing so.

ChrisThrasher:
> > What else do we need to do before we can release v2.6?
> 
> I guess it's good enough. Other things can be done/fixed in v2.6.x and breaking changes in 3.x.x, I guess

Sounds good. What is the release process? Is it as simple as cutting a tag, updating the CMakeLists.txt project version, then creating a new release in GitHub? If so then I'm happy to do that myself.

oprypin:
This pull request just drops support for *SFML* 2.5. Which is a significant loss that should have strong justification. Unfortunately the justification written in the description doesn't seem sufficient to me. We really don't have to update everything in lockstep. Releases of *imgui-sfml* at many points supported several versions. *SFML* 2.5 here was more of an exception that it was the only supported version for some time. And the fact that the latest release of both *sfml* and *imgui-sfml* was named "2.5" was actually also a coincidence.

*imgui-sfml* received several improvements, and we really shouldn't arbitrarily gatekeep those improvements just to *SFML* 2.6. So if this change is decided to merged, a release of *imgui-sfml* should be made *before* that.

oprypin:
This is a good pull request for *when* we decide to drop SFML 2.5 support, but as I said I think we really don't need to do that at the moment. We should just add both 2.5 and 2.6 to CI.

PR #246, opened 2023-07-30
Add and fix compiler warnings

ChrisThrasher:
Silencing the OpenGL deprecation warnings was my first goal since they were cluttering up the console output. While I was in there I added a quick a few lines to add the most basic GCC and Clang warnings. I see a few other PRs are touching on compiler warnings as well so this helps lay the groundwork to add (and fix) more warnings in the future.

I also kept another commit that fixes a little typo I found a few days ago.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ if (IMGUI_SFML_FIND_SFML)
   find_package(SFML 2.5 COMPONENTS graphics system window)
 
   if(NOT SFML_FOUND)
-    message(FATAL_ERROR "SFML 2 directory not found. Set SFML_DIR to directory where SFML was built (or one which ccontains SFMLConfig.cmake)")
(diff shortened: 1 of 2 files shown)


oprypin:
Hold up- I didn't think this through so well. This means that if in the future the compilers start spewing more warnings, people will not be able to build the project?

ChrisThrasher:
> This means that if in the future the compilers start spewing more warnings, people will not be able to build the project?

Theoretically yes, but in practice not likely with this basic set of warnings. Regardless it still may be a good idea to add a cache variable to disable `-Werror`. For now let's focus on adding and fixing more warnings because the #1 way to address potential compiler warnings is to aggressively find and fix them. We need to enable and fix more warnings until we approach the set of warnings that SFML uses. That will require some extra stuff to make sure we don't apply any warnings to the ImGui source code that we're compiling.

PR #248, opened 2023-08-02
Revert enabling "warnings as errors" in the default build

oprypin:
* Following up and reverting part of https://github.com/SFML/imgui-sfml/pull/246

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,12 +131,6 @@ target_compile_definitions(ImGui-SFML
     IMGUI_USER_CONFIG="${IMGUI_SFML_CONFIG_NAME}"
 )
 
-if(MSVC)
-  target_compile_options(ImGui-SFML PRIVATE /W3)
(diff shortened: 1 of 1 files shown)


ChrisThrasher:
This is overkill. Remove `-Werror` if you're worried about that. Better yet add a new option like SFML does that lets users disable warnings as errors. It's a very bad idea to entirely remove compiler warnings because it's increasing the chances of bugs or other code smells getting in.

ChrisThrasher:
https://github.com/SFML/imgui-sfml/commit/7da7dfe10ee27c48878195940a5355eed60e3104

Here's the fix you actually want. Our list of warnings is only going to grow so it's good to have this option for those who for whatever reason are seeing a warning that isn't already caught in CI.

oprypin:
I want to make that release happen, and take no risks in it.

ChrisThrasher:
See https://github.com/SFML/imgui-sfml/pull/251

PR #249, opened 2023-08-04
Simplify and expand CI pipeline

ChrisThrasher:
This new layout for CI drastically reduces the amount of boilerplate and makes it easy to expand support to more compilers.


oprypin:
Nice, I'll definitely take a look as soon as I find time

oprypin (approved):
All great stuff! Indeed then we should add SFML 2.5 and 3.0

eliasdaler:
Thanks a lot, this looks *much* better :+1:

PR #250, opened 2023-08-06
Fix new SFML 3 build error

ChrisThrasher:
See https://github.com/SFML/SFML/pull/2637

If we merge #249 then we can pretty easily add SFML 3 jobs to CI to continuously ensure SFML 3 support is maintained. Then as long as we run CI on a regular basis we will find these issues.


ChrisThrasher:
https://github.com/SFML/imgui-sfml/actions/runs/5774413453

I've got SFML 3 working in CI. It required this fix and another.

PR #251, opened 2023-08-08
Reenable compiler warnings

ChrisThrasher:
Reverts #248. Compiler warnings a critical part of quickly writing safe and correct code. It's imperative we use this. This PR adds compiler warnings that are disabled by default and only enforced as hard errors in CI. This way it does not pose a risk of 3rd parties failing to build our code due to `/WX` or `-Werror` turning a warning into an error.


Bambo-Borris (approved):
I'm shocked that these were removed. As an imgui-sfml power user, I would feel better knowing they're enabled.

PR #255, opened 2023-08-21
Fix finding SFML 3 and add SFML 3 to CI pipeline

ChrisThrasher:
Currently in `master` it's not possible to find an installation of SFML 3. Using SFML 3 requires that someone already builds SFML 3 in their build tree and thus can't depend on a preexisting installation.

I then added SFML 3 to CI to ensure SFML 3 support is constantly maintained.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ concurrency:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


ChrisThrasher:
Some recent SFML 3 changes broke the SFML build again.

PR #266, opened 2023-12-06
Added definitions for ImageButton with explicit ID

Sinan-Karakaya:
Hello!

I've added support for the new definitions of ImGui::ImageButton that support explicitely setting the id of the widget.
This feature was [added in ImGui v1.89](https://github.com/ocornut/imgui/issues/5533) and wasn't implemented in ImGui-SFML as far as I was aware.

Please give any feedback if something is wrong. Thanks :)

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -739,6 +739,15 @@ bool ImageButton(const sf::Texture& texture, const sf::Vector2f& size, const int
                               framePadding, toImColor(bgColor), toImColor(tintColor));
 }
 
+IMGUI_SFML_API bool ImageButton(const char* id, const sf::Texture& texture,
+                                const sf::Vector2f& size,
+                                const sf::Color& bgColor, const sf::Color& tintColor) {
+    ImTextureID textureID = convertGLTextureHandleToImTextureID(texture.getNativeHandle());
+
+    return ImGui::ImageButton(id, textureID, ImVec2(size.x, size.y), ImVec2(0, 0), ImVec2(1, 1),
+                              toImColor(bgColor), toImColor(tintColor));
+}
+
 /////////////// Image Button Overloads for sf::RenderTexture
 
 bool ImageButton(const sf::RenderTexture& texture, const int framePadding, const sf::Color& bgColor,
@@ -755,6 +764,18 @@ bool ImageButton(const sf::RenderTexture& texture, const sf::Vector2f& size, con
                               ImVec2(1, 0), // flipped vertically, because textures in
                                             // sf::RenderTexture are stored this way
                               framePadding, toImColor(bgColor), toImColor(tintColor));
+}
+
+IMGUI_SFML_API bool ImageButton(const char* id, const sf::RenderTexture& texture,
+                                const sf::Vector2f& size,
+                                const sf::Color& bgColor, const sf::Color& tintColor) {
+    ImTextureID textureID =
+        convertGLTextureHandleToImTextureID(texture.getTexture().getNativeHandle());
+
+    return ImGui::ImageButton(id, textureID, ImVec2(size.x, size.y), ImVec2(0, 1),
+                              ImVec2(1, 0), // flipped vertically, because textures in
+                                            // sf::RenderTexture are stored this way
+                              toImColor(bgColor), toImColor(tintColor));
 }
 
 /////////////// Image Button Overloads for sf::Sprite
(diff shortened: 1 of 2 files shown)


Sinan-Karakaya:
> This will require we raise the minimum ImGui requirement from 1.87 to 1.89. @eliasdaler Thoughts on raising that requirement?

For what's it's worth, it seems like the old definitions are still available in 1.89, although they are now marked as [deprecated](https://github.com/ocornut/imgui/commit/4a2ae06ca4996ca478b3e5e2f8f08e3b90ed2fe4).

ChrisThrasher:
Are there certain package managers still shipping <1.89 that we would break support with if we raised this requirement?

Sinan-Karakaya:
> Search the repo for "1.87". You should find another reference to that version in a build script.

Found it and fixed it in CMakeLists.txt

> Are there certain package managers still shipping <1.89 that we would break support with if we raised this requirement?

I would have to check, I mostly use FetchContent with ImGui-SFML so I'm not sure about that

CMakeLists.txt
```diff
@@ -32,13 +32,14 @@ endif()
 # ImGui does not provide native support for CMakeLists, workaround for now to have
 # users specify IMGUI_DIR. Waiting for this PR to get merged...
 #    https://github.com/ocornut/imgui/pull/1713
+set(IMGUI_DIR "${CMAKE_SOURCE_DIR}/imgui")
```
ChrisThrasher, line 35: Please remove this

ChrisThrasher:
> Search the repo for "1.87". You should find another reference to that version in a build script.

Search again. There are more references to 1.87.

ChrisThrasher:
Thanks for cleaning that up.

I'm going to give Elias time to weigh in one whether he likes increasing this minimum requirement.

eliasdaler:
> I'm going to give Elias time to weigh in one whether he likes increasing this minimum requirement.

Let's do it. 1.89 was released more than a year ago and Dear ImGui is a fast moving library which many users are expected to keep up with since it rarely breaks its API.

eliasdaler:
Shouldn't be overloads without "const char* id" be removed? It seems like you can't call ImGui::ImageButton without passing a label/id now.

imgui-SFML.cpp
```diff
@@ -785,6 +806,25 @@ bool ImageButton(const sf::Sprite& sprite, const sf::Vector2f& size, const int f
                               toImColor(bgColor), toImColor(tintColor));
 }
```
eliasdaler, line 807: For example, this should be removed. I'm not sure how it compiles now

imgui-SFML.cpp
```diff
@@ -785,6 +806,25 @@ bool ImageButton(const sf::Sprite& sprite, const sf::Vector2f& size, const int f
                               toImColor(bgColor), toImColor(tintColor));
```
eliasdaler, line 806: For example, I'm not sure how this compiles with ImGui 1.89+

Sinan-Karakaya: This should compile fine [considering this](https://github.com/ocornut/imgui/blob/v1.89/imgui_widgets.cpp#L1083)

eliasdaler:
> > Shouldn't overloads without `const char* id` be removed? It seems like you can't call ImGui::ImageButton without passing a label/id since 1.89.
> 
> They can be kept, as those definitions are still available in 1.89. The only difference is that they are now marked as deprecated.

I see. Let's just drop these overloads - it's better than copy-paste and we'll need to get rid of them eventually.

Sinan-Karakaya:
> > > Shouldn't overloads without `const char* id` be removed? It seems like you can't call ImGui::ImageButton without passing a label/id since 1.89.
> > 
> > 
> > They can be kept, as those definitions are still available in 1.89. The only difference is that they are now marked as deprecated.
> 
> I see. Let's just drop these overloads - it's better than copy-paste and we'll need to get rid of them eventually.

Just dropped them in last commit

eliasdaler:
Much better.
@ChrisThrasher, please look at it again and feel free to merge.

eliasdaler (approved):

imgui-SFML.cpp
```diff
@@ -726,48 +726,33 @@ void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Color&
...
 /////////////// Image Button Overloads for sf::Texture
 
-bool ImageButton(const sf::Texture& texture, const int framePadding, const sf::Color& bgColor,
-                 const sf::Color& tintColor) {
-    return ImageButton(texture, sf::Vector2f(texture.getSize()), framePadding, bgColor, tintColor);
-}
-
-bool ImageButton(const sf::Texture& texture, const sf::Vector2f& size, const int framePadding,
-                 const sf::Color& bgColor, const sf::Color& tintColor) {
+IMGUI_SFML_API bool ImageButton(const char* id, const sf::Texture& texture,
```
ChrisThrasher, line 729:
```suggestion
bool ImageButton(const char* id, const sf::Texture& texture,
```

imgui-SFML.cpp
```diff
@@ -726,48 +726,33 @@ void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Color&
...
 /////////////// Image Button Overloads for sf::RenderTexture
 
-bool ImageButton(const sf::RenderTexture& texture, const int framePadding, const sf::Color& bgColor,
-                 const sf::Color& tintColor) {
-    return ImageButton(texture, sf::Vector2f(texture.getSize()), framePadding, bgColor, tintColor);
-}
-
-bool ImageButton(const sf::RenderTexture& texture, const sf::Vector2f& size, const int framePadding,
-                 const sf::Color& bgColor, const sf::Color& tintColor) {
+IMGUI_SFML_API bool ImageButton(const char* id, const sf::RenderTexture& texture,
```
ChrisThrasher, line 740:
```suggestion
bool ImageButton(const char* id, const sf::RenderTexture& texture,
```
I don't think you need these macros in the .cpp file. You only need them in the header.

imgui-SFML.cpp
```diff
@@ -726,48 +726,33 @@ void Image(const sf::Sprite& sprite, const sf::Vector2f& size, const sf::Color&
...
-bool ImageButton(const sf::Sprite& sprite, const int framePadding, const sf::Color& bgColor,
-                 const sf::Color& tintColor) {
-    const sf::FloatRect spriteSize = sprite.getGlobalBounds();
-    return ImageButton(sprite, sf::Vector2f(spriteSize.width, spriteSize.height), framePadding,
-                       bgColor, tintColor);
-}
-
-bool ImageButton(const sf::Sprite& sprite, const sf::Vector2f& size, const int framePadding,
-                 const sf::Color& bgColor, const sf::Color& tintColor) {
+IMGUI_SFML_API bool ImageButton(const char* id, const sf::Sprite& sprite, const sf::Vector2f& size,
```
ChrisThrasher, line 754:
```suggestion
bool ImageButton(const char* id, const sf::Sprite& sprite, const sf::Vector2f& size,
```

ChrisThrasher (changes requested):

imgui-SFML.cpp
Sinan-Karakaya: You're right, just fixed it

ChrisThrasher (approved):

PR #267, opened 2023-12-07
Use conversion constructor

ChrisThrasher:
We can simplify imgui-SFML.cpp a decent amount by using the `ImVec` constructor we define in imconfig-SFML.h.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -241,7 +241,7 @@ bool Init(sf::Window& window, const sf::Vector2f& displaySize, bool loadDefaultF
     initDefaultJoystickMapping();
 
     // init rendering
-    io.DisplaySize = ImVec2(displaySize.x, displaySize.y);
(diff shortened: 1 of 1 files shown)


eliasdaler:
Here's a problem with this - it makes `imconfig-SFML.h` required and it defines `IM_VEC2_CLASS_EXTRA` which not all users might want.

This might be especially problematic if they want to have their own `IM_VEC2_CLASS_EXTRA` which defines conversion from their vector types to ImVec2/4 and imgui-sfml always defining it might lead to some problems, especially if using imgui-sfml as a shared library.

ChrisThrasher:
If I'm understanding correct, this code results in a hard compilation failure if someone changes the value of `IMGUI_SFML_USE_DEFAULT_CONFIG` to point to a different header?

ChrisThrasher:
See #268 

A better solution would be to add a `toImVec` function in the same anonymous namespace as `toImColor`.

PR #275, opened 2024-05-02
Fix forward declaration of `sf::Event`

ChrisThrasher:
https://github.com/SFML/SFML/pull/2766 changed `sf::Event` from a `struct` to a `class`.


Bambo-Borris (approved):
Sneaky `struct` keyword was hiding, thanks for the quick fix!

PR #287, opened 2024-07-20
Fix for removed Event Inheritance

AlexAUT:
Event inheritance was removed in SFML3. I can also split the lambda into two if that is preferred over the auto parameter.

### DIFF
--- a/examples/minimal/main.cpp
+++ b/examples/minimal/main.cpp
@@ -15,9 +15,9 @@ int main() {
     sf::Clock deltaClock;
     while (window.isOpen()) {
         while (const auto event = window.pollEvent()) {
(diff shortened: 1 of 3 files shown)


ChrisThrasher:
Looks like we need to update the events to handle how `pollEvent` now returns a `std::optional<sf::Event>` rather than just an `sf::Event`. Would you mind fixing that real quick?

ChrisThrasher (approved):

AlexAUT:
You can also close #285 after this one is merged.

ChrisThrasher:
> You can also close #285 after this one is merged.

lol I forgot to merge my own PR. My bad.

PR #298, opened 2024-10-15
Fix ImTextureID type

pifopi:
Hello !

I just wanted to report an issue with latest ImGUI. More specifically
https://github.com/ocornut/imgui/commit/92b94980c69ff3d3d7f5dc30c1c19abfb72db47e

I do know that my current solution means older ImGUI version wouldn't work anymore so I'm wondering what would be the best solution in your opinion (Check the type of ImTextureID and assign it accordingly ? Force the type of ImTextureID to be void* ?)

Thanks in advance

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -894,7 +894,7 @@ SpriteTextureData getSpriteTextureData(const sf::Sprite& sprite)
 
 ImTextureID convertGLTextureHandleToImTextureID(GLuint glTextureHandle)
 {
-    ImTextureID textureID = nullptr;
+    ImTextureID textureID = 0;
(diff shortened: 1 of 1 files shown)


peytonk132:
I had the same problem and I made it a reinterpret_cast so it becomes `reinterpret_cast<ImTextureID>(nullptr);`

pifopi:
After thinking a bit more, using the {} constructor would make it work no matter what the underlying type is (no matter if the user pick a different type, is on a previous version or the current version). Also, it avoids relying on define

ChrisThrasher:
> After thinking a bit more, using the {} constructor would make it work no matter what the underlying type is (no matter if the user pick a different type, is on a previous version or the current version). Also, it avoids relying on define

Sounds good to me

PR #300, opened 2024-10-24
Update imgui-SFML.cpp

KROIA:
casting nullptr to ImTextureID to prevent MSVC Err. C2440


ChrisThrasher:
https://github.com/SFML/imgui-sfml/pull/298

Can use the same fix we already merged into `master`? I'd rather not use a C-style cast.

ChrisThrasher:
I tried to update your PR but accidentally just pushed straight to 2.6.x. Anyways, this should be fixed now.

Issue #301, opened 2024-10-26
Compilation with IMGUI_DISABLE_OBSOLETE_FUNCTIONS


Legulysse:
Hi !

I am working on a project based on SFML 2.6.x, currently using ImGui 1.87 and ImGui-SFML 2.6.x.
Everything compiles with IMGUI_DISABLE_OBSOLETE_FUNCTIONS.

I tried upgrading ImGui to its latest version, but there are compilation errors with imgui-sfml when using this flag, though it's mostly due to renames.

I took the time to fix the build locally and now I have a working build with the latest imgui release (v1.91.4) with IMGUI_DISABLE_OBSOLETE_FUNCTIONS flag active.
The only problem with this is that depending on the ImGui version, building with those updates will break the build, and it would require to upgrade the minimum version supported (or use some version #define blocks).
From my tests, it should compile when using v1.91.1 as minimum version.

Would you be interested in a PR including the code changes?
If that is the case, what is the recommended course of action ? If I am not mistaken, I should fork, create a branch, push my commits, then propose a PR from my fork/fix-branch into this repo/2.6.x-branch ? I could also propose a dedicated branch if it needs some iterations/tests from others.

The problem is that I use premake instead of cmake, so I will not be able to update the cmake configuration tied to those modifications.
I also dont have the ability to test that gamepad navigation is behaving correctly.

ChrisThrasher:
> IMGUI_DISABLE_OBSOLETE_FUNCTIONS

What is the purpose of this option?

> but there are compilation errors with imgui-sfml when using this flag, though it's mostly due to renames.

Will you please share the compilation errors?

What's with these API breaks within the same major version of ImGui? Does the library not use Semantic Versioning? We had to fix another API break just earlier this week.

Legulysse:
This option is used to disable all the deprecation workarounds used by ImGui, when the library is making breaking API changes.
The intention is to allow projects to use newer versions of ImGui without the need to immediately update their code to accommodate with those API changes, though there is no guaranty as to how long those workarounds will stay in the API.

This flag forces those workarounds away.
It is recommended by the ImGui author to try compiling with this flag enforced from time to time to avoid being caught by surprise with too many breaking API changes.

I personally prefer to keep this flag active to force me to update things as soon as I upgrade my ImGui version.
There is nothing mandatory with this though, and I could probably only apply this flag on my own codebase and ignore it when building imgui-sfml.

Here is a list of the things that are deprecated when using the flag :
[...]

PR #302, opened 2024-10-26
Remove use of obsolete ImGui functions

Legulysse:
This PR contains migration commits to allow imgui-sfml to be compiled with the flag IMGUI_DISABLE_OBSOLETE_FUNCTIONS.
I put them in the order of imgui updates for easier integration if needed.

Those commits will force the project to raise the minimum imgui version supported, so it may not be welcome right now.
However, it can be useful for people willing to enforce this compilation flag on their projects.
I am not familiar with cmake and thus I am unable to tweak/test the version settings on this part.

Please feel free to use those commits as you see fit, either by cherry-picking them at a later time, or by merging them into a different branch.

This PR is tied to this ticket : https://github.com/SFML/imgui-sfml/issues/301

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       with:
         repository: ocornut/imgui
         path: imgui
-        ref: v1.89
+        ref: v1.91.1
 
     - name: Checkout SFML
       uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


ChrisThrasher:
Now that CI is checking for the use of obsolete functions, if we want to remove more obsolete functions then what we're really doing is raising the minimum ImGui requirement. That's a valid thing to consider, I just want to make sure we're on the same page about what it entails.

For now I'm going to close this PR because as it stands I'm not yet sure I want to raise the ImGui requirement for ImGui-SFML 2. ImGui-SFML 3 on the other hand certainly ought to raise that requirement. I'm not sure what version to pick but I think something newer is in order.

PR #303, opened 2024-10-31
Remove use of obsolete ImGui functions - v1.89

Legulysse:
This PR contains a migration commit to allow imgui-sfml to be compiled with the flag IMGUI_DISABLE_OBSOLETE_FUNCTIONS.
Target imgui version : 1.89

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ project(imgui_sfml VERSION 2.6 LANGUAGES CXX)
 option(IMGUI_SFML_FIND_SFML "Use find_package to find SFML" ON)
 option(IMGUI_SFML_IMGUI_DEMO "Build imgui_demo.cpp" ON)
 option(IMGUI_SFML_ENABLE_WARNINGS "Enable compiler warnings" OFF)
+option(IMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS "Disable obsolete ImGui functions" OFF)
 
 # If you want to use your own user config when compiling ImGui, please set the following variables
 # For example, if you have your config in /path/to/dir/with/config/myconfig.h, set the variables as follows:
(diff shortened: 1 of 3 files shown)


Legulysse:
Nice ! 
To be clear, this update will not impact people already using 1.89, it will only make it available for people using the stricter compilation with the deprecation flag.

CMakeLists.txt
```diff
@@ -84,6 +85,9 @@ if(BUILD_SHARED_LIBS)
   set_target_properties(ImGui-SFML PROPERTIES DEFINE_SYMBOL "IMGUI_SFML_EXPORTS")
   set_target_properties(ImGui-SFML PROPERTIES DEBUG_POSTFIX "_d")
 endif()
+if(IMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS)
+  target_compile_definitions(ImGui-SFML PRIVATE IMGUI_DISABLE_OBSOLETE_FUNCTIONS)
```
ChrisThrasher, line 89: Does this need to be made `PUBLIC`?

ChrisThrasher (approved):
Decided to make `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` `PUBLIC` so that consumers of the `ImGui-SFML` CMake target also get that defined. This means that when they compile Imgui.h they don't have access to functions that were not compiled into their build of ImGui-SFML.

Legulysse:
Seems safer this way to me ! And it allows the library and its users to move forward alongside imgui's future versions.
(you left a typo, using "obsolute" instead of "obsolete" in the cmake doc)

ChrisThrasher:
Thanks for your help on this. I'll get this merged into `master` as well. I'm still not sure what to do about future ImGui versions which make even more things obsolete but we can hopefully figure something out. In the meantime I hope this makes your life a little easier.

Legulysse:
@ChrisThrasher Thank you for your work !
I guess I should close the other PR I made ?
Dont hesitate if you want to discuss those concerns further on the associated issue, or if you want me to provide PRs for those other updates later on.

PR #308, opened 2024-12-06
Test with 2 versions of ImGui

ChrisThrasher:


ChrisThrasher:
@Legulysse I decided I want to add two version of ImGui to CI, doubling the job count. That should be fine since our CI pipeline. Is this our best strategy? Test with the minimum version w/o deprecated interfaces and the latest version w/ deprecated interfaces?

Legulysse:
@ChrisThrasher It seems like a really good idea to check both the latest supported version, and the minimum supported version !  

Since imgui is maintained in a way that provides as much backwards-compatibility as possible, having the minimum version tested with the deprecation flag could be seen as unneeded, but I believe it's a good incentive to help moving the code forward, as long as it does not put excessive work on maintaining the latest supported version 🤔

PR #310, opened 2024-12-23
Updated obsoleted functions

fgallegosalido:
The members `ImGuiIO::SetClipboardTextFn` and `ImGuiIO::GetClipboardTextFn` have been obsoleted. This pull request uses the new members in `ImGuiPlatformIO`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,8 +37,8 @@ jobs:
         - { name: Release }
         - { name: Debug }
         imgui:
-        - { version: 1.89, flags: -DIMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS=ON }
-        - { version: 1.91.5 }
+        - { version: 1.91.1 }
+        - { version: 1.91.6, flags: -DIMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS=ON }
 
     steps:
     - name: Get CMake and Ninja
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,8 +37,8 @@ jobs:
         - { name: Release }
         - { name: Debug }
         imgui:
-        - { version: 1.91.1 }
-        - { version: 1.91.6, flags: -DIMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS=ON }
+        - { version: 1.91.1, flags: -DIMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS=ON }
+        - { version: 1.91.6 }
 
     steps:
     - name: Get CMake and Ninja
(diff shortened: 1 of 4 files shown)


ChrisThrasher:
Your solution breaks compatibility with ImGui 1.89. What newer version of ImGui to you propose that ImGui-SFML 3 use?

fgallegosalido:
It seems that the earliest that supports this is 1.91.1, so I would say that one, although I would consider using the latest (1.91.6), which is the recommended way to go.

ChrisThrasher:
How do you acquire ImGui for use with ImGui-SFML? A local git clone? A package manager? CMake FetchContent?

fgallegosalido:
Local git clone.

ChrisThrasher:
How does ImGui-SFML stand to benefit from using a newer version of ImGui? Are there better features or anything else we stand to benefit from?

fgallegosalido:
This is the recommended way to go according to the developer: [https://github.com/ocornut/imgui/blob/master/docs/FAQ.md#q-which-version-should-i-get]().

ChrisThrasher:
> It seems that the earliest that supports this is 1.91.1, so I would say that one

I think I'm cool with this. We still haven't raised the ImGui requirement for ImGui-SFML 3 yet but I know that needs to happen at some point. This seems like a reasonable choice. Not necessarily sure I want to pick the latest ImGui release so 1.91.1 sounds good for now.

.github/workflows/ci.yml
```diff
@@ -37,8 +37,8 @@ jobs:
         - { name: Release }
         - { name: Debug }
         imgui:
-        - { version: 1.89, flags: -DIMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS=ON }
-        - { version: 1.91.5 }
+        - { version: 1.91.1 }
+        - { version: 1.91.6, flags: -DIMGUI_SFML_DISABLE_OBSOLETE_FUNCTIONS=ON }
```
ChrisThrasher, line 41: Why did you swap this around?

fgallegosalido: Doesn't it make sense to disable obsoleted functions in the latest tested version? I don't think it makes much sense to do it in the minimum supported one, which we know is not going to trigger any new failures

ChrisThrasher: https://github.com/SFML/imgui-sfml/issues/301

Please read through this issue to get caught up on past conversions we've had on the matter. Disabling obsolete functions in the newest version will break compatibility with the oldest version.

fgallegosalido: Right, I see. Let me revert the swap. Do I leave the latest version change or just add 1.91.6 without removing 1.91.5?

ChrisThrasher: You can still update 1.91.5 to 1.91.6. That's something I plan on periodically doing as ImGui updates to make sure nothing about ImGui breaks the ImGui-SFML build.

ChrisThrasher (approved):

Issue #326, opened 2025-07-07
imgui 1.92.0 ImTextureID


Luksx:
with 1.92.0 imgui release there was a change in ImTextureID to ImTextureRef
Since this change, it does not build.

> All API functions taking a ImTextureID parameter are now taking a 'ImTextureRef`:
ImTextureRef is a small composite structure which may be constructed from a ImTextureID (or constructed from a ImTextureData* which represent a texture which will generally be ready by the time of rendering).

ChrisThrasher:
Ugh, ImGui keeps breaking their API at a fast pace. Is there any alternate solution that also lets ImGui-SFML support 1.91 or does supporting 1.92 necessarily require dropping support for older versions? I don’t want ImGui-SFML to be constantly raising it’s ImGui requirement to chase down these API breakages if we can help it.

Luksx:
I just managed to get it working with replacing line 956 
`assert(io.Fonts->TexID != (ImTextureID) nullptr);`
with
`assert(io.Fonts->TexRef._TexData != nullptr); `

Actually no clue why does it work xd

TiphaineLAURENT:
I would guess it is possible (but probably not wanted) to use a macro to test the imgui version : https://github.com/ocornut/imgui/blob/v1.92.0/imgui.h#L31

```cpp
#if IMGUI_VERSION_NUM >= 19200
    assert(io.Fonts->TexRef._TexData != nullptr); // You forgot to create and set font texture
#else
    assert(io.Fonts->TexID != (ImTextureID) nullptr); // You forgot to create and set font texture
#endif
```

Romi-Brooks:
> I just managed to get it working with replacing line 956 with`assert(io.Fonts->TexID != (ImTextureID) nullptr);` `assert(io.Fonts->TexRef._TexData != nullptr); `
> 
> Actually no clue why does it work xd

In ImGui v1.92.2b, an error occurs in imgui-SFML.cpp at line 1043:
```cpp
const GLuint textureHandle = convertImTextureIDToGLTextureHandle(pcmd->TextureId);
```

Said that "'const struct ImDrawCmd' has no member named 'TextureId'"

use “const GLuint textureHandle = convertImTextureIDToGLTextureHandle(pcmd->GetTexID());”, for me ,it's works fine.

Perhaps we should revert to v1.91.1.

fgallegosalido:
[...]

Issue #319, opened 2025-04-15
bug: couldn't compile with ImGui 1.91.9


CosminPerRam:
Upon using `v3.0` and ImGui `1.91.9` I encounter these compilation errors:
```
D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-src\imgui-SFML.cpp(779,12): error C2661: 'ImGui::Image': no overloaded function takes 6 arguments [D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-build\ImGui-SFML.vcxproj]
  D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-src\imgui-SFML.cpp(779,12):
  while trying to match the argument list '(ImTextureID, ImVec2, ImVec2, ImVec2, ImColor, ImColor)'

D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-src\imgui-SFML.cpp(792,12): error C2661: 'ImGui::Image': no overloaded function takes 6 arguments [D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-build\ImGui-SFML.vcxproj]
  D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-src\imgui-SFML.cpp(792,12):
  while trying to match the argument list '(ImTextureID, ImVec2, ImVec2, ImVec2, ImColor, ImColor)'

D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-src\imgui-SFML.cpp(813,12): error C2661: 'ImGui::Image': no overloaded function takes 6 arguments [D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-build\ImGui-SFML.vcxproj]
  D:\GitHub\StylishVisualizer\build\_deps\imgui-sfml-src\imgui-SFML.cpp(813,12):
  while trying to match the argument list '(ImTextureID, ImVec2, ImVec2, ImVec2, ImColor, ImColor)'
```
Code references (in order of error messages):
https://github.com/SFML/imgui-sfml/blob/070d6ab9aa98c997b1fa846175d3faf93de3866e/imgui-SFML.cpp#L779
https://github.com/SFML/imgui-sfml/blob/070d6ab9aa98c997b1fa846175d3faf93de3866e/imgui-SFML.cpp#L792-L800
https://github.com/SFML/imgui-sfml/blob/070d6ab9aa98c997b1fa846175d3faf93de3866e/imgui-SFML.cpp#L813

[...]

PR #330, opened 2025-09-11
feat: support for 1.92.0

majvax:
This is primarily a proof of concept rather than a production-ready implementation. I am more than happy to collaborate with the maintainers to refine this update according to your preferences and project standards.

**Please note:** This PR should **not** be merged in its current state. It lacks proper testing, and the code was written during my class, so there may be performance or memory issues that I have overlooked.

My work is based on the official documentation and the SDL3 renderer backend.

I welcome any suggestions or guidance on how to improve this contribution. 

Reference:
- [sdl3 1.92.0 update commit](https://github.com/ocornut/imgui/commit/e538883a20bd5896f33f1ee2717b171a593347d3)
- [official doc to update backend for 1.92.0](https://github.com/ocornut/imgui/commit/e538883a20bd5896f33f1ee2717b171a593347d3)

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -1,4 +1,5 @@
 #include "imgui-SFML.h"
+#include <cstdint>
 #include <imgui.h>
 
 #include <SFML/Config.hpp>
@@ -33,6 +34,8 @@
 #include <SFML/System/NativeActivity.hpp>
 #include <android/native_activity.h>
 #include <jni.h>
+
+
 
 int openKeyboardIME()
 {
@@ -313,6 +316,7 @@ bool Init(sf::Window& window, const sf::Vector2f& displaySize, bool loadDefaultF
     io.BackendFlags |= ImGuiBackendFlags_HasGamepad;
     io.BackendFlags |= ImGuiBackendFlags_HasMouseCursors;
     io.BackendFlags |= ImGuiBackendFlags_HasSetMousePos;
+    io.BackendFlags |= ImGuiBackendFlags_RendererHasTextures;
     io.BackendPlatformName = "imgui_impl_sfml";
 
     s_currWindowCtx->joystickId = getConnectedJoystickId();
@@ -337,13 +341,6 @@ bool Init(sf::Window& window, const sf::Vector2f& displaySize, bool loadDefaultF
     loadMouseCursor(ImGuiMouseCursor_ResizeNESW, sf::Cursor::Type::SizeBottomLeftTopRight);
     loadMouseCursor(ImGuiMouseCursor_ResizeNWSE, sf::Cursor::Type::SizeTopLeftBottomRight);
     loadMouseCursor(ImGuiMouseCursor_Hand, sf::Cursor::Type::Hand);
-
-    if (loadDefaultFont)
-    {
-        // this will load default font automatically
-        // No need to call AddDefaultFont
-        return UpdateFontTexture();
-    }
 
     return true;
 }
@@ -589,6 +586,15 @@ void Shutdown(const sf::Window& window)
 {
     const bool needReplacement = (s_currWindowCtx->window->getNativeHandle() == window.getNativeHandle());
 
+    for (ImTextureData* tex : ImGui::GetPlatformIO().Textures)
+    {
+        if (tex->RefCount == 1)
+        {
+            tex->SetStatus(ImTextureStatus_WantDestroy);
+            UpdateFontTexture(tex);
+        }
+    }
+
     // remove window's context
     auto found = std::find_if(s_windowContexts.begin(),
                               s_windowContexts.end(),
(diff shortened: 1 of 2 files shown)


majvax:
The second commit is based of #319. As noted in the [v1.91.9](https://github.com/ocornut/imgui/releases/tag/v1.91.9) changelog, parameters `tint_col` and `border_col` have been remove from `ImGui::Image`. The signature is deprecated and remplaced by `ImGui::ImageWithBg`.

After some testing, I found that the minimum supported version for these changes is v1.92.0. The new interface for textures introduced in this version breaks compatibility with earlier releases, so support for versions prior to 1.92.0 is not feasible. The good news is that these updates are compatible up to v1.92.2b, and ImGui can now be compiled with IMGUI_DISABLE_OBSOLETE_FUNCTIONS enabled.

majvax:
I believe upgrading to 1.92.0 is necessary, as ImGui will require this update moving forward. As quoted from the doc:

> "In order to move forward and take advantage of all new features, support for `ImGuiBackendFlags_RendererHasTextures` will likely be REQUIRED for all backends before June 2026"

majvax:
@ChrisThrasher What do you think of this?

ChrisThrasher:
So you’re saying that supporting ImGui 1.92 means all prior versions become unusable? We have to raise the minimum from 1.91.1 to 1.92?

majvax:
That’s basically what I’m saying, yes. Certainly, we could use macros to provide version-specific code, but I don’t think that fits the modern C++ approach of SFML.

ChrisThrasher:
Does ImGui provide macros for detecting the current version?

majvax:
I didn't look at it myself but trusting [this comment](https://github.com/SFML/imgui-sfml/issues/326#issuecomment-3048546498) I would say yes.

mikademus:
Would it be helpful to stage the updates rather than try to get all fixes in at once? It seems that the most urgent change is the texture changes. ImGui-SFLM overloads ImGui::Image and ImGui::ImageButton for sf::Texture, sf::Sprite and sf::RenderTexture, and I think basically need updates to the texture ID semantic? This seems separate from the font revisions, and a manageable high-reward step.

Caerind:
Hello!
I'm using a different version of imgui-sfml, but thank you a lot your implementation was a good starting point!
Though I found some issues that needs to be addressed : 
- In UpdateFontTexture, you are doing '=' instead of '==' in all the conditions !
- In UpdateFontTexture, I would remove the else and use 'tex->Status == ImTextureStatus_WantDestroy', this is probably not required, but more robust
- And in the same branch, you are doing the SetTexID to invalid before getting it. So you never actually destroy the concerned texture
- UpdateFontTexture should be renamed UpdateTexture probably
- We should remove the "fontTexture" and its access in the .h
- Also, I removed the usage of the unordered_map by storing a raw pointer to a sf::Texture* I manage directly and store into the tex->BackendUserData

Caerind:
I also had to make some other changes to support more fonts, I made the commit in a fork:
https://github.com/SFML/imgui-sfml/commit/862d4595afbd0b055bc494a583290453ff082911

majvax:
Hey, first I'm not really into ImGui programming nor SFML itself and I don't have the experience to actually come with a fix. As I said this code was wrotten during class and is poorly written. I just serve as a poc of the support for 1.92.0.
Addressing the different point made in your message
- great catch for the conditions ! I don't really know how could I wrote that nor how my compiler/build tool didn't catch it, I believe that before AI slop is human slop haha
- using `else if` instead of `else` is as you said way more robust.
- nice catch I must admit I quite laugh at this snippet of code while reviewing it
- I didn't understand what you meant by `fontTexture` as I wrote this code a while ago and didn't find anything related into the code except the `s_currWindowCtx->fontTexture`. Could you precise more what you meant here?
- for the unordered_map, I found the tex->BackendUserData but never fix it since 1.92.0 is not something planned.

To sum up I needed 1.92.0 for a project that is now finished. For the sake of open source I will update the code when I have the time to even if 1.92.0 support isn't planned.

majvax:
> I also had to make some other changes to support more fonts, I made the commit in a fork: [862d459](https://github.com/SFML/imgui-sfml/commit/862d4595afbd0b055bc494a583290453ff082911)

Thank you for your contribution. If you’re interested, I can close my PR and you can open a new one based on your fork/commit.

Caerind:
Hey !
I should have precised it in my post, but I'm using ImGui 1.92.5 & SFML 2.6 (for a stupid reason, but don't ask why ahah), so I have some custom changes on imgui-sfml.

> I didn't understand what you meant by fontTexture as I wrote this code a while ago and didn't find anything related into the code except the s_currWindowCtx->fontTexture. Could you precise more what you meant here?

Yes exactly! That fontTexture is not needed anymore, so it should be removed, and with that the GetFontTexture function.

I made a PR which is here : https://github.com/SFML/imgui-sfml/pull/336
(Also I think it is currently not working because I have some Texture changes that are not on the main branch, but I'll update that very soon)

And thank you again, your PR helped me for a good start!

PR #332, opened 2025-12-13
Fix Android define to standard

Caerind:
__ANDROID__ is the standard define for Android platform. There is still the USE_JNI that is user-defined, so the code won't be enabled for people not want the feature on Android.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -27,7 +27,7 @@
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #endif
 
-#ifdef __ANDROID__
+#ifdef SFML_SYSTEM_ANDROID
 #ifdef USE_JNI
 
 #include <SFML/System/NativeActivity.hpp>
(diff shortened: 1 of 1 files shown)


Caerind:
There are also changes that I needed on my side (but on imgui-sfml v2.6.1):
- Commented **glPolygonMode** in SetupRenderState
- "ifndef-ed" **last_polygon_mode** in RenderDrawLists
- "ifndef-ed" **glPopAttrib** the 2 **glPolygonMode** in RenderDrawLists

(Edit: Maybe that first commented line could be ifdef instead 🤣)

eXpl0it3r:
> **ANDROID** is the standard define for Android platform. There is still the USE_JNI that is user-defined, so the code won't be enabled for people not want the feature on Android.

Any source to quote that `__ANDROID__` is the right define?



> There are also changes that I needed on my side (but on imgui-sfml v2.6.1):
> 
>     * Commented **glPolygonMode** in SetupRenderState
> 
>     * "ifndef-ed" **last_polygon_mode** in RenderDrawLists
> 
>     * "ifndef-ed" **glPopAttrib** the 2 **glPolygonMode** in RenderDrawLists
> 
> 
> (Edit: Maybe that first commented line could be ifdef instead 🤣)

Is this in any way related to the given PR?

Caerind:
I think the safest would be this one: https://github.com/SFML/SFML/blob/ca65abe717cc731b5ee309d25c1291508e163795/include/SFML/Config.hpp#L74

(So I updated the PR to use it directly)


For the comments, it's not directly related to the PR. These are fixes that I needed on Android, but I don't know if they are valid. Just mentioning if that can help anyone using imgui-SFML on Android and having similar issues.

PR #335, opened 2026-01-30
Fix symbol export/import on windows

lapinozz:
Fixed the DLLs macros so the shared build would work with MSVC

IMGUI_SFML_SHARED_LIB should be defined public in shared build, so that when you link against the lib you get dll import
IMGUI_SFML_EXPORTS is defined private so that the lib itself compiles with the dll export

IMGUI_API and IMGUI_SFML_API should both be the same since the source for both is compiled directly into the lib so I simplified the code and defined IMGUI_API  as IMGUI_SFML_API


CMakeLists.txt
```diff
@@ -77,8 +77,8 @@ if(WIN32 AND MINGW)
   target_link_libraries(ImGui-SFML PUBLIC imm32)
 endif()
 if(BUILD_SHARED_LIBS)
-  target_compile_definitions(ImGui-SFML PRIVATE IMGUI_SFML_SHARED_LIB)
-  set_target_properties(ImGui-SFML PROPERTIES DEFINE_SYMBOL "IMGUI_SFML_EXPORTS")
+  target_compile_definitions(ImGui-SFML PUBLIC IMGUI_SFML_SHARED_LIB)
+  target_compile_definitions(ImGui-SFML PRIVATE IMGUI_SFML_EXPORTS)
```
ChrisThrasher: Let’s combine these into one `target_compile_definitions` call

imgui-SFML_export.h
```diff
@@ -5,21 +5,17 @@
...
 #ifdef IMGUI_SFML_EXPORTS
 #define IMGUI_SFML_API __declspec(dllexport)
-#define IMGUI_API      __declspec(dllexport)
 #else
 #define IMGUI_SFML_API __declspec(dllimport)
-#define IMGUI_API      __declspec(dllexport)
 #endif
 #elif __GNUC__ >= 4
 #define IMGUI_SFML_API __attribute__((visibility("default")))
-#define IMGUI_API      __attribute__((visibility("default")))
```
ChrisThrasher, line 15: Let’s remove the `else` clause for GCC 4 and prior while we’re touching this file. No need for that anymore.

ChrisThrasher: We want to remove the code when GCC < 4. You removed the code for when GCC is greater than or equal to 4

ChrisThrasher (changes requested):

CMakeLists.txt
```diff
@@ -77,8 +77,8 @@ if(WIN32 AND MINGW)
   target_link_libraries(ImGui-SFML PUBLIC imm32)
 endif()
 if(BUILD_SHARED_LIBS)
-  target_compile_definitions(ImGui-SFML PRIVATE IMGUI_SFML_SHARED_LIB)
-  set_target_properties(ImGui-SFML PROPERTIES DEFINE_SYMBOL "IMGUI_SFML_EXPORTS")
+  target_compile_definitions(ImGui-SFML PUBLIC IMGUI_SFML_SHARED_LIB
+                                        PRIVATE IMGUI_SFML_EXPORTS)
```
eXpl0it3r, line 81: SFML also uses `set_target_properties` + `DEFINE_SYMBOL`: https://github.com/SFML/SFML/blob/51196acddb879ee1a276ba1a39c1fec590114abf/cmake/Macros.cmake#L99C5-L99C84

What speaks for a private definition or is this equivalent?

lapinozz: target_compile_definitions is the modern way of doing it

set_target_properties is the equivalent of a private target_compile_definitions 
the symbol should be defined when compiling the files in the target but not in the files importing/linking the target

eXpl0it3r (approved):

PR #336, opened 2026-02-06
Backend: Renderer textures for 1.92.5

Caerind:
Today I had a look at this PR: https://github.com/SFML/imgui-sfml/pull/330
To start to implement it on my custom imgui-sfml version.
The previous PR has some issues that I fixed.

I used the ImGui backend page : https://github.com/ocornut/imgui/blob/master/docs/BACKENDS.md
And also the OpenGL impl : https://github.com/ocornut/imgui/commit/dbb91a5

And adapted what needed to be done to support SFML and the current backend properly.

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -243,9 +243,6 @@ struct WindowContext
     const sf::Window* window;
     ImGuiContext*     imContext{ImGui::CreateContext()};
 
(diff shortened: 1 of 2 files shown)


cristianglezm:
~~is this for SFML 3.x or 2.6.x?~~

edited: it is for 2.6.x, you will need to change the target of the pull request, as I think master is for SFML 3.x

PR #339, opened 2026-03-18
Fix clip space rect condition

lapinozz:
It seems like the code would check that the right and bottom of the rect were above 0 instead of checking the actual width and height. This would create log spam in certain circumstances.


eXpl0it3r (approved):
Might want to consider cleaning up the commit history and/or to squash merge the PR.

Issue #340, opened 2026-04-02
ImGui-SFML is not compatible with latest ImGui version (> 1.92.0)


jacogasp:
Hello everyone,

from the open issues and PRs I see that there is already some movement from this side, but it is not clear to me if the effort in supporting latest ImGui version (> 1.92.0) is still alive or not.

I bootstrapped an empty project based on CMake and Vcpkg, adding `imgui-sfml` as the only dependency, here you can find the source code https://github.com/jacogasp/imgui-sfml-test/tree/main.

The example provided by this project's readme crashes with

```
Assertion failed: (io.Fonts->Fonts.Size > 0), function Update, file imgui-SFML.cpp, line 548.
[1]    12340 abort      ./build/imgui-sfml-test
```

This is due to the new ImGui scalable fonts system, as many have already reported.

Even though in the documentation is clearly written 
> Default font is loaded if you don't pass false in `ImGui::SFML::Init.`

it is not true anymore.

The only way to start the example is to add the snippet

```
auto& io = ImGui::GetIO();
io.Fonts->AddFontDefault();
```

just below the window initialization `ImGui::SFML::Init(window)`.

So far so good, except that fonts are corrupted and mostly unreadable as shown in figure (running on macOS).

<img width="752" height="624" alt="Image" src="https://github.com/user-attachments/assets/0a8f85fd-b4de-465a-9413-c337c27a9919" />

I tried with some external .tff fonts and the behavior is the very same.

Since in vcpkg there is no easy way to specify sub-dependencies version, this project is currently broken at the moment, at least for who manages dependencies via package managers like vcpkg.

Edit: The topic is covered here https://github.com/SFML/imgui-sfml/issues/326, https://github.com/SFML/imgui-sfml/pull/330 and https://github.com/SFML/imgui-sfml/pull/336.

lapinozz:
As you saw there's already a couple PRs and issues on this subject. I'm planning on taking a closer look at those PRs and starting a branch to support 1.92 soon

jacogasp:
Hello @lapinozz, 

after a couple of days of investigation I found the reason of the broken font rendering of the picture above.

I took the opportunity to open a new PR (https://github.com/SFML/imgui-sfml/pull/341) inspired by https://github.com/SFML/imgui-sfml/pull/336 and https://github.com/SFML/imgui-sfml/pull/330.

Long story short:
[...]

PR #341, opened 2026-04-05
Add support for ImGui v1.92.0

jacogasp:
With version v1.92.0, ImGui introduced the RenderHasTexture backend flag to enable backend to render fonts as raw textures. The current ImGui-SFML backend implementation does not support this feature and minimal examples don't run because of failing assertions.

This commit provide support for the new RenderHasTexture backend flag while supporting previous version using the IMGUI_VERSION_NUM macro.

Inspired by https://github.com/SFML/imgui-sfml/pull/336 and https://github.com/SFML/imgui-sfml/pull/330

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Dependencies
 -----
 
 * [SFML](https://github.com/SFML/SFML) >= 3.0.0
-* [Dear ImGui](https://github.com/ocornut/imgui) >= 1.91.1, < 1.92.0
+* [Dear ImGui](https://github.com/ocornut/imgui) >= 1.91.1
 
 Contributing
 -----
@@ -141,6 +141,18 @@ int main() {
 
 Fonts how-to
 ---
+
+Starting with ImGui v1.92.0, default font is automatically loaded.
+
+* You can load your fonts like this:
+
+```cpp
+IO.Fonts->Clear(); // clear fonts if you loaded some before (even if only default one was loaded)
+IO.Fonts->AddFontFromFileTTF("font1.ttf", 8.f);
+IO.Fonts->AddFontFromFileTTF("font2.ttf", 12.f);
+```
+
+**ImGui < v1.92.0** (deprecated)
 
 Default font is loaded if you don't pass `false` in `ImGui::SFML::Init`. Call `ImGui::SFML::Init(window, false);` if you don't want default font to be loaded.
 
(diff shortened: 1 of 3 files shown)


lapinozz:
I don't think I like the idea of supporting two API for imgui-sfml itself at the same time.
I'd rather we find a way to keep the same API or just deprecate the old version if we have to

lapinozz:
Maybe there's a way to emulate the new API for imgui < 1.92
Since the new version always loads the default font we could get rid of that parameter and always load the font in the old version too
Maybe there's a way to detect stale font texture so the user doesn't have to manually call the UpdateFontTexture function?

lapinozz:
Here's what I came up with. It works with ImGui < 1.92 but it has the same API and behaviour as ImGui >= 1.92
https://github.com/SFML/imgui-sfml/compare/master...lapinozz:imgui-sfml:imgui-1.92

jacogasp:
Hello @lapinozz,

thank you for having a look to the PR. Yea, also to me is not very elegant the usage of the macro to check ImGui version, it makes it a little cumbersome in the sake of backward compatibility. Actually I don't have an opinion on dropping entirely support for ImGui < 1.92.
Anyway, your changes to support all versions look good to me.

milipili:
Just to inform that at least on MacOS it is currently not working great.
When using latest sfml-imgui on MacOS via vcpkg, fonts are no longer working and it crashes when quitting.
The `UpdateFontTexture()` requires changes of course (actually simply removed)

However those version in vcpkg work flawlessly
- imgui:arm64-osx@1.91.9
- imgui-sfml:arm64-osx@3.0#1

lapinozz:
I think the version on vcpkg has some code changed applied, hopefully once everything is figured out those should be removed 

The fonts are not working even with ImGui 1.91?

jacogasp:
> Just to inform that at least on MacOS it is currently not working great.
> 
> When using latest sfml-imgui on MacOS via vcpkg, fonts are no longer working and it crashes when quitting.
> 
> The `UpdateFontTexture()` requires changes of course (actually simply removed)
> 
> 
> 
> However those version in vcpkg work flawlessly
> 
> - imgui:arm64-osx@1.91.9
> 
> - imgui-sfml:arm64-osx@3.0#1

Yes I confirm that the vcpkg port does not work because of a patch that insert a bug in UpdateFontTexture. For making this work, I added a custom overlay that does not contain the patch and points to my fork. I added some details [here](https://github.com/SFML/imgui-sfml/issues/340).

I tested on macOS arm.

PR #348, opened 2026-05-24
Fix GL state saving and application

### LINKED ISSUE #347: Mixing SFML code with imgui-sfml broken on mobile
Currently mixing SFML rendering with imgui-sfml seems to not work properly together.

On Android, I'm getting the following message spammed:
`<glDrawArrays:4884>: glDrawArrays is called with VERTEX_ARRAY client state disabled!`

After further investigation it seems like rendering from imgui-sfml indeed disables the vertex arrays, color arrays and texture arrays:

Rendering happens here: https://github.com/SFML/imgui-sfml/blob/master/imgui-SFML.cpp#L573

There is a call to `resetGLStates()` which reset these vertex/color/texture arrays to be enabled. After that there is a call to the main rendering function which after rendering disables these arrays: https://github.com/SFML/imgui-sfml/blob/master/imgui-SFML.cpp#L1061

From SFML code (https://github.com/SFML/SFML/blob/master/src/SFML/Graphics/RenderTarget.cpp#L523) only `resetGLStates()` resets the vertex arrays, color arrays and texture arrays back to enabled which does not seem to be called after imgui-sfml rendering. 

This breaks the SFML rendering? I'm not sure why I don't have any problems with this on desktop?

Indeed I was able to fix this locally by moving the `resetGLStates()` to be after rendering like this:
```
    target.pushGLStates();
    ImGui::Render();
    RenderDrawLists(ImGui::GetDrawData());
    target.popGLStates();
    target.resetGLStates();
```
I'm unsure if that now may break something else.

lapinozz:
Corrected some of the GL states we save and restore 
Should fix https://github.com/SFML/imgui-sfml/issues/347

### DIFF
--- a/imgui-SFML.cpp
+++ b/imgui-SFML.cpp
@@ -969,19 +969,19 @@ void RenderDrawLists(ImDrawData* draw_data)
     GLint last_blend_dst;
     glGetIntegerv(GL_BLEND_SRC, &last_blend_dst);
 
-    const auto last_blend          = glIsEnabled(GL_BLEND);
-    const auto last_cull_face      = glIsEnabled(GL_CULL_FACE);
-    const auto last_depth_test     = glIsEnabled(GL_DEPTH_TEST);
-    const auto last_stencil_test   = glIsEnabled(GL_STENCIL_TEST);
-    const auto last_lighting       = glIsEnabled(GL_LIGHTING);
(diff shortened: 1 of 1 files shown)


Zombieschannel (approved):
Seems to have fixed the issue, thanks

PR #343, opened 2026-04-20
Modernize input handling

lapinozz:
Updated the input handling to fix some bugs and to make it more similar to imgui's own backends (mostly looked at the SDL one)

- Use events for touch instead of saving the state and setting it every frame
- Pass events to imgui even if we are not in focus, if we receive the events we should pass them
- Use SFML's function to show the virtual keyboard instead of manually managing it for Android

Fixes https://github.com/SFML/imgui-sfml/issues/337, https://github.com/SFML/imgui-sfml/issues/338 and https://github.com/SFML/imgui-sfml/issues/88

PR #351, opened 2026-08-12
Fix Windows CI

TealSeer:
Changes in the `windows-2025` image caused Windows CI jobs to start failing, there was two problems to be fixed:
- MSVC builds started complaining that `nmake` couldn't be found, this was fixed by adding a build step that sets the environment variables you'd get if you ran the VS Developer Command Prompt
- ClangCL builds were complaining that the NMake generator doesn't support toolset specification, fixed by changing the generator to Ninja and explicitly passing `DCMAKE_C/CXX_COMPILER` flags instead of `-T`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         platform:
         - { name: Windows MSVC,    os: windows-2025 }
-        - { name: Windows ClangCL, os: windows-2025, flags: -T ClangCL }
+        - { name: Windows ClangCL, os: windows-2025, flags: -GNinja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl }
         - { name: Windows Clang,   os: windows-2025, flags: -GNinja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ }
         - { name: Linux GCC,       os: ubuntu-22.04, flags: -GNinja }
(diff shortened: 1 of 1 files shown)


eXpl0it3r:
Both these issues can be fixed by updating CMake to a newer version on Windows.

The issue is that the selected version of CMake doesn't know about the VS 2026 generator and thus fallsback to the NMake generator.

The issue arises with the new image, as VS 2026 was introduced replacing VS 2022.

I recently fixed this for SFML itself.

TealSeer:
> Both these issues can be fixed by updating CMake to a newer version on Windows.
> 
> The issue is that the selected version of CMake doesn't know about the VS 2026 generator and thus fallsback to the NMake generator.
> 
> The issue arises with the new image, as VS 2026 was introduced replacing VS 2022.
> 
> I recently fixed this for SFML itself.

I'm currently experimenting with this on another branch, just bumping the CMake version for Windows runners caused the build to fail, seemingly because the FreeType version depended on CMake 3, that required bumping SFML itself from 3.0 to 3.1 which turned it from what I thought would be a simple fix into a Whole Thing.

EDIT: After playing with it some, I can get CI to pass after bumping CMake and SFML, but it requires #343 to be merged (the build errors are warnings about deprecation being treated as errors). So maybe something to look into after that's done, I'll attach the patch file for posterity.
[bump.patch](https://github.com/user-attachments/files/30998660/bump.patch)

lapinozz:
@TealSeer I applied your changes to the repo, thanks for the help!

